PR #6, opened 2024-06-14
Update handling of translations for errors

j-maynard:
We found we were passing raw error strings back to the frontend with no knoweldge of which language the user was using.  This PR updates the language handling providing error strings in both English and Welsh for anyone trying to consume the Backend directly and prvodies the trasnlation tags and parameters for consuming in the frontend.

This PR also contains some code improvements.


easternbloc (approved):
LGTM

PR #7, opened 2024-08-05
Implement Title and Datafile Preview

j-maynard:
This PR implemetns using the title from the start of the journey and the preview table parts of the journey.  This gives us the first part of the user journey in working software.  We switch to using Datalake storage for preview files to using blob stoage.


src/controllers/csv-processor.ts
```diff
@@ -250,3 +315,96 @@ export const processCSV = async (dataset: Dataset, page: number, size: number):
...
+                dataset_id: dataset.id
+            };
+        }
+        const buff = await blobStoageService.readFile(`${datafile.id}.csv`);
+
+        const dataArray: Array<Array<string>> = (await parse(buff, {
+            delimiter: ','
+        }).toArray()) as string[][];
+        const csvheaders = dataArray.shift();
+        const total_pages = Math.ceil(dataArray.length / size);
```
andyjpb, line 356: This has the same issue as a previous commit where we read in the whole file only to throw most of it away. So a paginated file amplifies the load on the service by the number of pages instead of reducing it by the number of pages.
If we need this information then we can could cache it in a rows_hint field in the database. We'd probably also need some index data that told us the offset in bytes of each page.

src/controllers/csv-processor.ts
```diff
@@ -250,3 +315,96 @@ export const processCSV = async (dataset: Dataset, page: number, size: number):
...
+        const errors = validateParams(page, total_pages, size);
+        if (errors.length > 0) {
+            return {
+                success: false,
+                errors,
+                dataset_id: dataset.id
+            };
+        }
+
+        const csvdata = paginate(dataArray, page, size);
```
andyjpb, line 366: The pagination stuff needs to evolve to work with seekable streams rather than fully materialised arrays.

src/dtos/user-dto.ts
```diff
@@ -0,0 +1,19 @@
+import { User } from '../entity/user';
+
+export class UserDto {
```
andyjpb, line 3: We need to identify and use the persistent ID to identify objects of this class.
We need to specify whether the login ID / eMail is either changeable or re-assignable for each IDP we want to work with.

src/dtos/user-dto.ts
```diff
@@ -0,0 +1,19 @@
...
+    id: string;
+    oidcId: string;
```
andyjpb, line 5: Presumably this is the persistent ID?
If so, please add a comment to this effect.
If not, add a different comment to different effect. ;-)

src/dtos/user-dto.ts
```diff
@@ -0,0 +1,19 @@
...
+    provider: string;
```
andyjpb, line 6: Presumably this uniquely identifies the IDP?
If so, please add a comment to this effect.
If not, add a different comment to different effect. ;-)

src/dtos/user-dto.ts
```diff
@@ -0,0 +1,19 @@
...
+    name: string;
```
andyjpb, line 7: In the general case this can change at every login. We need to decide if we care about a history of changes or if there is any user facing thing we need to do when it happens.

src/dtos/user-dto.ts
```diff
@@ -0,0 +1,19 @@
...
+    email: string;
```
andyjpb, line 8: In the general case this can change at every login. We need to decide if we care about a history of changes or if there is any user facing thing we need to do when it happens.

A change here may indicate a change or it may indicate a reassignment. We should understand our persistent ID so that we can make a change here an entirely user facing issue.

What do we need this for and therefore what do we do when/if we don't get it for a particular user?

src/route/dataset-route.ts
```diff
@@ -73,9 +84,10 @@ apiRoute.post('/', upload.single('csv'), async (req: Request, res: Response) =>
         res.json(err);
         return;
     }
-    const dataset = Dataset.createDataset(internalName, 'BetaUser');
+    const dataset = Dataset.createDataset(title, 'BetaUser');
     const saved_dataset_record = await dataset.save();
-    const uploadDTO = await uploadCSV(req.file?.buffer, saved_dataset_record);
+    saved_dataset_record.addTitleByString(title, lang);
+    const uploadDTO = await uploadCSVToBlobStorage(req.file?.buffer, saved_dataset_record);
```
andyjpb, line 90: When we receive file data via HTTP POST it gets encoded in the Request Body.
In these cases we should be able to tell the HTTP Server implementation to only read the Request Header and then pass the body to us as an open file descriptor. Then we can decode the parameters one-by-one and when it gets to the (potentially large) file data, pass the FD (or a derivative) onto lower layers so that we can process the data page-by-page straight from the incoming TCP socket to the outgoing buffer (disk or TCP).
Then we don't have to worry about how much data the user sends us other than to check the parameter size and ensure we have enough space in the backing store. i.e. we don't need more than a few pages (32K?) of RAM for any sized upload.

src/route/dataset-route.ts
```diff
@@ -101,6 +113,11 @@ apiRoute.get('/:dataset', async (req, res) => {
         res.json({ message: 'Dataset not found... file is null or undefined' });
         return;
     }
+    if (isValidUUID(datasetID) === false) {
```
andyjpb, line 116: This check is unnecessary if we check for the existence of datasetID as a database object and that query is properly protected against injections.

src/route/dataset-route.ts
```diff
@@ -101,6 +113,11 @@ apiRoute.get('/:dataset', async (req, res) => {
...
+        res.status(404);
+        res.json({ message: 'Dataset not found...File ID is not Valid.' });
```
andyjpb, line 118: Is this really a fileID? We shouldn't expose filenames directly to users, even if we do check them for form. We should only expose the IDs of database objects.
Should this say something like "Malformed dataset id"?

src/route/dataset-route.ts
```diff
@@ -131,6 +148,11 @@ apiRoute.get('/:dataset/csv', async (req, res) => {
         res.json({ message: 'Dataset not found... Dataset ID not found in Database' });
         return;
     }
+    if (isValidUUID(datasetID) === false) {
```
andyjpb, line 151: Copypasta? Why do we do this check again?

andyjpb: Ah... it's another route handler.
Perhaps factor this stuff out as it appears in another handler again below.

src/route/dataset-route.ts
```diff
@@ -163,6 +185,11 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
         res.json({ message: 'Dataset not found... You must specify a dataset ID' });
         return;
     }
+    if (isValidUUID(datasetID) === false) {
```
andyjpb, line 188: Factor out as above?

src/route/dataset-route.ts
```diff
@@ -174,13 +201,47 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
...
 
+apiRoute.get('/:dataset/preview', async (req, res) => {
+    const datasetID = req.params.dataset;
+    if (datasetID === undefined || datasetID === null) {
+        res.status(404);
+        res.json({ message: 'Dataset not found... You must specify a dataset ID' });
+        return;
+    }
+    if (isValidUUID(datasetID) === false) {
+        res.status(404);
```
andyjpb, line 212: There's another Response code for Bad Request.

src/route/dataset-route.ts
```diff
@@ -174,13 +201,47 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
     });
 });
 
+apiRoute.get('/:dataset/preview', async (req, res) => {
+    const datasetID = req.params.dataset;
+    if (datasetID === undefined || datasetID === null) {
+        res.status(404);
```
andyjpb, line 207: There's another Response code for Bad Request.

src/route/dataset-route.ts
```diff
@@ -174,13 +201,47 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
...
+        res.json({ message: 'Dataset not found...File ID is not Valid.' });
+        return;
+    }
+    const dataset = await Dataset.findOneBy({ id: datasetID });
+    if (dataset === undefined || dataset === null) {
+        res.status(404);
+        res.json({ message: 'Dataset not found... Dataset ID not found in Database' });
+        return;
+    }
```
andyjpb, line 221: This looks like code similar to above. Perhaps factor this out too?

src/route/dataset-route.ts
```diff
@@ -174,13 +201,47 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
...
+    const page_number_str: string = req.query.page_number || req.body?.page_number;
+    const page_size_str: string = req.query.page_size || req.body?.page_size;
+    const page_number: number = Number.parseInt(page_number_str, 10) || 1;
+    const page_size: number = Number.parseInt(page_size_str, 10) || DEFAULT_PAGE_SIZE;
+    const processedCSV = await processCSVFromBlobStorage(dataset, page_number, page_size);
+    if (!processedCSV.success) {
+        res.status(400);
```
andyjpb, line 228: Should this be 500? Any error in processCSVFromBlobStorage() sounds like a server fault, not a client fault as all client params have been validated by this time.

src/route/dataset-route.ts
```diff
@@ -174,13 +201,47 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
...
+    }
+    res.json(processedCSV);
```
andyjpb, line 230: As per earlier comment, rework this to be a streamed response rather than batched through an array in memory.

src/route/dataset-route.ts
```diff
@@ -174,13 +201,47 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
...
+});
+
 apiRoute.get('/:dataset/view', async (req, res) => {
     const datasetID = req.params.dataset;
     if (datasetID === undefined || datasetID === null) {
         res.status(404);
```
andyjpb, line 236: There's another Response code for Bad Request.

src/route/dataset-route.ts
```diff
@@ -174,13 +201,47 @@ apiRoute.get('/:dataset/xlsx', async (req, res) => {
...
         res.json({ message: 'Dataset not found... You must specify a dataset ID' });
         return;
     }
+    if (isValidUUID(datasetID) === false) {
+        res.status(404);
```
andyjpb, line 241: There's another Response code for Bad Request.

src/route/dataset-route.ts
```diff
@@ -191,7 +252,7 @@ apiRoute.get('/:dataset/view', async (req, res) => {
     const page_size_str: string = req.query.page_size || req.body?.page_size;
     const page_number: number = Number.parseInt(page_number_str, 10) || 1;
     const page_size: number = Number.parseInt(page_size_str, 10) || DEFAULT_PAGE_SIZE;
-    const processedCSV = await processCSV(dataset, page_number, page_size);
+    const processedCSV = await processCSVFromDatalake(dataset, page_number, page_size);
```
andyjpb, line 255: As per earlier comment, rework this to be a streamed response rather than batched through an array in memory.

src/route/user-route.ts
```diff
@@ -0,0 +1,41 @@
...
+userRoute.get('/:id', async (req, res) => {
+    if (!req.params.id) {
+        logger.error(`Reequest parameter id is missing`);
+        res.status(400).send();
+        return;
+    }
+    const userID = req.params.id;
+    const user = await User.findOneBy({ id: userID });
+    if (!user) {
+        logger.error(`User with id ${req.params.id} not found`);
```
andyjpb, line 19: log4j-style vulnerability due to unescaped injection of user controlled variable.

src/route/user-route.ts
```diff
@@ -0,0 +1,41 @@
...
+    res.json(userDto);
+});
+
+userRoute.post('/', async (req, res) => {
+    const user = new User();
+    user.oidcId = req.body.oidcId;
+    user.provider = req.body.provider;
+    user.name = req.body.name;
+    user.email = req.body.email;
+    user.profile = req.body.profile;
```
andyjpb, line 33: All this data should come from the IDP.
If there's code to verify that in the OIDC library then we can rely on it but we need a comment here to explain it.
Otherwise we need to validate all of these things before putting them in the database.
The name and eMail, which will later be used in HTML Responses are particularly sensitive.

andyjpb: i.e. we need to ensure it's safe if injected by an untrusted attacker (and specify how much we trust the IDP and our OIDC client). Document (in a comment) what attribute checking has happened up to this point and then do whatever other checking we need in code.

src/route/user-route.ts
```diff
@@ -0,0 +1,41 @@
...
+    let usr = await User.findOneBy({ oidcId: user.oidcId });
+    if (usr === null || usr === undefined) {
+        usr = await user.save();
```
andyjpb, line 36: Do we need to set a flag for new users so that we can put them through some kind of introductory flow?

src/route/user-route.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (!user) {
+        logger.error(`User with id ${req.params.id} not found`);
+        res.status(404).send();
+        return;
+    }
+    const userDto = userToUserDTO(user);
+    res.json(userDto);
+});
+
+userRoute.post('/', async (req, res) => {
```
andyjpb, line 27: We probably need a specific endpoint for handling OIDC logins.
See comment at end of procedure below.

src/route/user-route.ts
```diff
@@ -0,0 +1,41 @@
...
+    user.provider = req.body.provider;
+    user.name = req.body.name;
+    user.email = req.body.email;
+    user.profile = req.body.profile;
+    let usr = await User.findOneBy({ oidcId: user.oidcId });
+    if (usr === null || usr === undefined) {
+        usr = await user.save();
+    }
+    const userDto = userToUserDTO(usr);
+    res.status(201).send();
```
andyjpb, line 39: > We probably need a specific endpoint for handling OIDC logins.

...so that we can extract the context we supplied at the beginning of the login journey (i.e. before we redirected them to the IDP) and then redirect them back to the part of our site that they started on.

package.json
```diff
@@ -55,6 +55,8 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
+    "@azure/identity": "^4.4.1",
```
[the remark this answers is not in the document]
andyjpb: Towards the end of the diff there is a lot of work on login that is not mentioned in the commit message.

andyjpb (changes requested):
Only a few comments but it was quite hard to review because lots of independent stuff at different layers of the stack were all in the same commit.

Commit messages don't mention login changes but they were in the diff.

src/controllers/blob-storage.ts
```diff
@@ -0,0 +1,102 @@
...
+        return uploadBlobResponse;
+    }
+
+    public async deleteFile(fileName: string) {
+        logger.warn(`Deleting file with file '${fileName}' from blob storage`);
+        const blockBlobClient = this.containerClient.getBlockBlobClient(fileName);
+        await blockBlobClient.delete();
+    }
+
+    public async listFiles() {
```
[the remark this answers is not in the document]
j-maynard: Its not currently used by our code but is implemented for completeness for covering the main things we might need to do with the API.  Further down the line we will almost certainly want some form of house keeping methods to be able to keep the temporary storage clear.

src/controllers/blob-storage.ts
```diff
@@ -0,0 +1,102 @@
...
+        const fileList: string[] = [];
+        for await (const blob of this.containerClient.listBlobsFlat()) {
```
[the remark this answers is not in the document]
j-maynard: So list isn't a streaming interface...  Its just an API response with a list of blobs held in the container.  This is just a simple interface for listing the contents of the blob container.

src/controllers/blob-storage.ts
```diff
@@ -0,0 +1,102 @@
...
+            fileList.push(blob.name);
+        }
+        return fileList;
+    }
+
+    public async readFile(fileName: string) {
+        const blockBlobClient = this.containerClient.getBlockBlobClient(fileName);
+        const downloadBlockBlobResponse = await blockBlobClient.download();
```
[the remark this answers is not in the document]
j-maynard: The API will handle streams and at a later date I'll be happy to reimplement using streams as we move towards more complete code.

src/controllers/blob-storage.ts
```diff
@@ -0,0 +1,102 @@
...
+            throw new Error('Failed to get readable stream body from download response.');
+        }
+
+        const chunks: Buffer[] = [];
+        for await (const chunk of readableStreamBody) {
+            if (chunk instanceof Buffer) chunks.push(chunk);
+            else chunks.push(Buffer.from(chunk));
+        }
+        return Buffer.concat(chunks);
+    }
```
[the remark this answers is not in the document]
j-maynard: One reads to a buffer rather than to a string.  I left the string implementation for debugging purposes currently.  As noted both implementations might go away if we move to streams.

package.json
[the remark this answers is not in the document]
j-maynard: This isn't that sort of identity... This about software making requests to get Identity from Microsoft Azure automatically.

src/controllers/blob-storage.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    public async uploadFile(fileName: string | undefined, fileContent: Buffer) {
+        if (fileName === undefined) {
+            throw new Error('File name is undefined');
+        }
+        if (fileContent === undefined) {
+            throw new Error('File name is undefined');
+        }
+        logger.debug(`Uploading file with file '${fileName}' to blob storage`);
```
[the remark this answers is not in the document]
andyjpb: If it doesn't care then it needs to take steps to protect itself from this vulnerability.
If we don't want to do that then we need to document the assumptions as part of the interface. i.e. this function should only be passed known-untainted data.

src/controllers/blob-storage.ts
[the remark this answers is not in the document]
andyjpb: Mark as "janitorial use only" or something so it doesn't end up being called by user facing code without further review?

src/controllers/blob-storage.ts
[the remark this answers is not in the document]
andyjpb: Why do you have to `await` it then?

package.json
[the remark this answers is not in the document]
andyjpb: Does anything in this patch rely on this or can we punt it to another commit/PR?

src/controllers/csv-processor.ts
```diff
@@ -101,16 +102,75 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
+    const datafile: Datafile | undefined = datafiles
+        .filter((filterfile: Datafile) => filterfile.draft === false)
+        .sort(
+            (first: Datafile, second: Datafile) =>
+                new Date(second.creationDate).getTime() - new Date(first.creationDate).getTime()
+        )
+        .shift();
+    const dto = await datasetToDatasetDTO(dataset);
+    if (datafile) {
+        try {
```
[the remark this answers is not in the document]
j-maynard: So I think the risk is small... It doesn't matter if the file ends up in both places so long as ultimately it ends up in the datalake for the creation of the "online cube".   Cleaning up is nice, but not essential to how the system should operate.

The client currently is really only intended to be the frontend service.  The user talks to the frontend, the frontent talks to the backend.  If they user breaks the request to the frontend this doesn't automatically break the request to the backend.

Additionally moving files from Azure Blob -> Backend Service (running in the same Azure tenancy) -> Azure Datalake should take milliseconds for all but the largest of files and by large I am thinking multi-gigabyte files.  I've yet to see evidence of files above about 20mb currently.

src/controllers/csv-processor.ts
```diff
@@ -101,16 +102,75 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
+                    ],
+                    tag: { name: 'errors.no_csv', params: {} }
+                }
+            ]
+        };
+    }
+};
+
+export const uploadCSVToBlobStorage = async (buff: Buffer, dataset: Dataset): Promise<UploadDTO | UploadErrDTO> => {
+    const blobStorageService = new BlobStorageService();
```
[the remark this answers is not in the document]
j-maynard: The try block can be found at line 170 in the `csv-processor.ts` file.  This BlobStorage class is a very thin layer over the Azure BlobStorage API.  its designed so it could be dropped for an entirely different API such as AWS S3 for example.

src/controllers/csv-processor.ts
```diff
@@ -101,16 +102,75 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
     const hash = createHash('sha256').update(buff).digest('hex');
     const datafile = Datafile.createDatafile(dataset, hash.toString(), 'BetaUser');
     const savedDataFile = await datafile.save();
     const dto = await datasetToDatasetDTO(dataset);
     if (buff) {
```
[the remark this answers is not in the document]
j-maynard: Well spotted... Fixed in a new commit.

src/controllers/csv-processor.ts
```diff
@@ -101,16 +102,75 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
         try {
```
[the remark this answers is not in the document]
j-maynard: Thats because this only deals with putting the file into Blob Storage.  We don't make any requests to the datalake at this point.  We are just storing the file the user supplied so that we can render it back to the user so they can preview what they uploaded.

As per the user journey the user then needs to confirm the file uploaded is the one they want to use as the initial fact table for the dataset.  If they confirm then we move the file from Blob storage to the Data Lake.

src/controllers/csv-processor.ts
```diff
@@ -250,3 +315,96 @@ export const processCSV = async (dataset: Dataset, page: number, size: number):
...
+                ],
+                dataset_id: dataset.id
+            };
+        }
+        const buff = await blobStoageService.readFile(`${datafile.id}.csv`);
+
+        const dataArray: Array<Array<string>> = (await parse(buff, {
+            delimiter: ','
+        }).toArray()) as string[][];
+        const csvheaders = dataArray.shift();
```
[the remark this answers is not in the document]
j-maynard: This will come when we reimplement the backend data model and we can capture that information, load it into the database and then get it back out again.  This will be iterated on soon.

src/controllers/csv-processor.ts
```diff
@@ -250,3 +315,96 @@ export const processCSV = async (dataset: Dataset, page: number, size: number):
...
+    try {
+        const datafiles = await dataset.datafiles;
+        const datafile: Datafile | undefined = datafiles
+            .filter((filterfile: Datafile) => filterfile.draft === true)
+            .sort(
+                (first: Datafile, second: Datafile) =>
+                    new Date(second.creationDate).getTime() - new Date(first.creationDate).getTime()
+            )
+            .shift();
+        if (datafile === undefined || datafile === null) {
```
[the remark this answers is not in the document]
j-maynard: Front the frontend the user is unable to upload anything that isn't a CSV.  We should check the mimetype at the point of initial ingestion on the backend API.  This isn't something we'll capture at the point.

src/controllers/csv-processor.ts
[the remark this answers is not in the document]
andyjpb: It's less about risk and more about robust software programming patterns. This kind of thing comes up everywhere so we should have a standard approach to handling it.
Additionally, if these things are not important enough to need to succeed then we should not be doing them in the HTTP Request/Response path.
I have some cubes here from SW2 that are several GB. See the Slack message from several months ago with the histogram that shows the distribution of SW2 cube sizes.

src/controllers/csv-processor.ts
j-maynard: Except pages aren't fixed in size... They can be any number from 5 to 500.  The current list is: 

- 5
- 10
- 20
- 50
- 100
- 250
- 500

We will have to paginate, we will have to paginate dynamically as well based on the users choices.  If its a big file we can't render everything directly to the frontend we will have to provide means to paginate that data until we are able to get it into the online cube.  Once its in the online cube we can query a specific number of lines and we can query a specific starting point.  CSV sadly doesn't give us the ability.

src/route/dataset-route.ts
j-maynard: This was implemented to prevent the app dying when asking for a dataset by ID which isn't a valid well formed UUID.  I didn't want to wrap the database request in try catch blocks when I can prevent malformed UUIDs from making it to the query line in first place and reject nice and early.  This will reduce load on the database and means we can kick-out bad requests as early as possible.

src/route/dataset-route.ts
j-maynard: Changed to `Dataset ID is not valid`

src/route/dataset-route.ts
j-maynard: Moved to a new function

src/route/dataset-route.ts
j-maynard: Moved to a new function

src/route/dataset-route.ts
j-maynard: Changed response code to `400`

src/route/dataset-route.ts
j-maynard: Changed response to `500`

src/route/dataset-route.ts
j-maynard: Changed response code to `500`

src/route/dataset-route.ts
j-maynard: Moved out to another function which standardised the response to `400`

src/route/dataset-route.ts
j-maynard: Changed to `400`

src/route/dataset-route.ts
j-maynard: I am prepared to do this in a later PR after I have reworked the backend following the new Data model.  I need to land this for show and tell and thats a bigger bit of work.

Dare I suggest we raise a Bug ticket in Jira and stick it in the next sprint.

src/controllers/csv-processor.ts
j-maynard: Possibly... We only need this code for the publisher journey though and only when we're directly handling the uploaded CSV files.  As soon as we're able to load this data in to the online cube much of this goes away as we'll just do SQL queries to get pages of the data.

src/controllers/csv-processor.ts
```diff
@@ -101,16 +102,75 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
+                        message: [
+                            { lang: ENGLISH, message: t('errors.move_to_datalake', { lng: ENGLISH }) },
+                            { lang: WELSH, message: t('errors.move_to_datalake', { lng: WELSH }) }
+                        ],
+                        tag: { name: 'errors.move_to_datalake', params: {} }
+                    }
+                ]
+            };
+        }
+    } else {
```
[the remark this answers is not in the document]
j-maynard: Its here because API requests are atomic... There is a possible state where there is no datafile attached to the Dataset.  This can occur because the user decided they were unhappy with the file they uploaded.  In this situation you end up in a position where there's a dataset but no datafile.  At this point a second user could make this request and there be no datafile to move.  So it can happen, it shouldn't happen but we have to code on the assumption that it will happen.

src/route/dataset-route.ts
j-maynard: I'm not 100% convinced we can do this using NodeJS and the layers of firewall, etc.  If you can provide me an example which proves me wrong I am happy to give it a look.  Ultimately NodeJS is its own Web Server with ExpressJS providing the framework for handling the post request.

andyjpb: I'm not sure about your API but other HTTP Request/Response APIs often provide a port object that can be used to read the body and there's a way to tell the framework to not process it.

https://expressjs.com/en/5x/api.html points me at http://expressjs.com/id/resources/middleware/body-parser.html which says it doesn't do multipart bodies and suggests some other modules.

Picking busboy (the first one) https://www.npmjs.com/package/busboy#readme, there's an example (the second one) and documentation: "file(< string >name, < Readable >stream, < object >info) - Emitted for each new file found. name contains the form field name. stream is a Readable stream containing the file's data."

...that shows how to parse multipart/form-data POST bodies (which is the only encoding that supports file uploads from browsers via HTML form submission).

Which module are we using for processing multipart/form-data?

andyjpb (approved):

PR #9, opened 2024-08-14
Implenet new datamodel on the backend

j-maynard:
Make all the changes to the backend to support the new database model.


src/entity2/import.ts
```diff
@@ -0,0 +1,30 @@
...
+    @JoinColumn({ name: 'csv_info' })
+    csvInfo: CsvInfo;
+
+    @Column({ type: 'varchar', length: 255 })
+    mime_type: string;
+
+    @Column({ type: 'varchar', length: 255 })
+    filename: string;
+
+    @OneToOne(() => Source, (source) => source.import)
```
andyjpb, line 28: Why is this OneToOne?
Each import represents a table of data.
Each time a column needs to be extracted a source will be specified to do it.

src/entity2/revision.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@Entity()
+export class RevisionEntity extends BaseEntity implements Revision {
+    @PrimaryGeneratedColumn('uuid')
+    id: string;
+
+    @Column({ type: 'int' })
+    revision_index: number;
+
+    @ManyToOne(() => Dataset, (dataset) => dataset.revisions)
```
andyjpb, line 28: ManyToOne?
For each revision there is only one dataset.
But for each dataset there can be many revisions.
From the POV of the `class RevisionEntity`, shouldn't it be OneToOne or perhaps OneToMany or have I got it backwards somehow? Afterall `(dataset) => dataset.revisions` could be reasonable but the syntax is bending my mind.

src/entity2/revision.ts
```diff
@@ -0,0 +1,55 @@
...
+    @JoinColumn({ name: 'dataset_id' })
+    dataset: Dataset;
+
+    @Column({ type: 'timestamp', default: () => 'CURRENT_TIMESTAMP' })
+    creation_date: Date;
+
+    @ManyToOne(() => RevisionEntity, { nullable: true })
```
andyjpb, line 35: There's only one "previous" revision for each revision but there could be zero, one or many "next" revisions for a particular revision.

src/entity2/revision.ts
```diff
@@ -0,0 +1,55 @@
...
+    @JoinColumn({ name: 'previous_revision_id' })
+    previous_revision: Revision;
+
+    @Column({ type: 'varchar', length: 255, nullable: true })
+    online_cube_filename: string;
```
andyjpb, line 40: I am not yet sure how many revisions of a particular dataset we can have online at a particular time (just the latest is the easiest and perhaps one extra for the current / most important draft), but yeah: it makes sense to store this in the revision rather than the dataset.

src/entity2/user.ts
```diff
@@ -0,0 +1,49 @@
+import { Entity, PrimaryGeneratedColumn, Column, BaseEntity } from 'typeorm';
+
+@Entity()
+export class User extends BaseEntity {
+    @PrimaryGeneratedColumn('uuid')
+    id: string;
+
+    @Column({ unique: true })
+    username: string;
```
andyjpb, line 9: I wouldn't bother having a username separate from an eMail address.

src/entity2/user.ts
```diff
@@ -0,0 +1,49 @@
...
+    @Column({ type: 'text', nullable: true })
+    refresh_token: string;
+
+    @Column({ type: 'text', nullable: true })
+    id_token: string;
+
+    @Column({ type: 'timestamp', nullable: true })
+    token_expiry: Date;
+
+    @Column({ nullable: true })
```
andyjpb, line 32: See "Falsehoods programmers believe about names".
Let's just have a single name field and ask the user to populate it with their preferred text the first time they log in.

src/entity2/user.ts
```diff
@@ -0,0 +1,49 @@
...
+    id: string;
+
+    @Column({ unique: true })
+    username: string;
+
+    @Column({ unique: true })
+    email: string;
+
+    @Column({ nullable: true, unique: true })
+    oidc_subject: string;
```
andyjpb, line 15: If this is NULL then I assume the object does not represent any kind of OIDC principal.
In that case do we need an alternative way to authenticate into the account such as a password? Or are such accounts only for system stuff such as batch jobs?

src/route/dataset-route.ts
```diff
@@ -39,72 +48,81 @@ function checkDatasetID(datasetID: string, res: Response): boolean {
...
+        res.json({ message: 'Error uploading file' });
+        return;
     }
+
+    // Everything looks good so far, let's create the dataset and revision records
+    const dataset = new Dataset();
+    dataset.creation_date = new Date();
+    // TODO change how we handle authentication to get the user on the Backend
+    dataset.created_by = 'Test User';
+    const saved_dataset_record = await dataset.save();
```
andyjpb, line 110: We don't need to await the DB transaction here as the object should have everything it needs for us to link to it in other objects later.

src/route/dataset-route.ts
```diff
@@ -39,72 +48,81 @@ function checkDatasetID(datasetID: string, res: Response): boolean {
...
+    const datasetInfo = new DatasetInfo();
+    datasetInfo.language = lang;
+    datasetInfo.title = title;
+    datasetInfo.dataset = saved_dataset_record;
```
andyjpb, line 114: This needs a flag to mark it as not-yet translated and which language is now the authoritative/current version of the text value.

src/route/dataset-route.ts
```diff
@@ -39,72 +48,81 @@ function checkDatasetID(datasetID: string, res: Response): boolean {
...
+    datasetInfo.save();
+    const revision = new DatasetRevision();
+    revision.dataset = saved_dataset_record;
+    revision.revision_index = 1;
```
andyjpb, line 118: The data model should have a way of allocating this itself.
It just needs to be monotonically increasing but it doesn't need to be consecutive or even dense. A single global counter will suffice.

src/route/dataset-route.ts
```diff
@@ -39,72 +48,81 @@ function checkDatasetID(datasetID: string, res: Response): boolean {
...
+    revision.creation_date = new Date();
+    // TODO change how we handle authentication to get the user on the Backend
```
andyjpb, line 120: Mention this in the commit message.

src/route/dataset-route.ts
```diff
@@ -39,72 +48,81 @@ function checkDatasetID(datasetID: string, res: Response): boolean {
...
+    revision.created_by = 'Test User';
+    const saved_revision_record = await revision.save();
```
andyjpb, line 122: There should be no need to await this here, as per previous comment.

src/route/dataset-route.ts
```diff
@@ -39,72 +48,81 @@ function checkDatasetID(datasetID: string, res: Response): boolean {
...
+    importRecord.revision = saved_revision_record;
+    importRecord.save();
+
```
andyjpb, line 125: Commit all the objects here in a single database transaction.
Currently there doesn't seem to be a transaction at all, so we need to add one.
The code as written also makes several round-trips to the database server which is not advisable given that we're using a remote server rather than something in-process.
Therefore, batch up all the queries as well. There should be some kind of batching / pipelining mode in the database driver.

andyjpb (changes requested):
A nice start! :-)

Just a few actions:
 + Separate out stream refactoring into its own commit.
 + Separate out structural model stuff into its own commit.
 + Separate out user data model stuff into its own commit.
 + Separate out routes and stuff into their own commit(s) that build on the groundwork from the structural model commit.
 + Provide a commit full of tests / example data to fill up the structural model: at least a few cubes with a few different dimensions, revisions and sources that demonstrate all the functionality so far.
 + Provide commit messages for all commits to explain rationale, design, progress so far, limitations, assumptions and general documentation for what we're thinking.

src/route/dataset-route.ts
j-maynard: We await the save so we can get the ID from the saved object.

src/controllers/blob-storage.ts
```diff
@@ -52,7 +53,7 @@ export class BlobStorageService {
 
         const blockBlobClient = this.containerClient.getBlockBlobClient(fileName);
 
-        const uploadBlobResponse = await blockBlobClient.upload(fileContent, fileContent.length);
+        const uploadBlobResponse = await blockBlobClient.uploadStream(fileContent, fileContent.readableLength);
```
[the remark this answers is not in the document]
j-maynard: We don't... We pass the stream to the Azure API to process how it wants to handle the stream... This is the point of hand off between code I wrote and code provided by Microsoft.  `fileContent` is a pointer to a `ReadableStream`

src/controllers/csv-processor.ts
```diff
@@ -18,6 +21,21 @@ export const MAX_PAGE_SIZE = 500;
 export const MIN_PAGE_SIZE = 5;
 export const DEFAULT_PAGE_SIZE = 100;
 
+function hashReadableStream(stream: Readable, algorithm: string = 'sha256'): Promise<string> {
```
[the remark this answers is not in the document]
j-maynard: At some stage we have to read the stream if you want hashes of files... In terms of order... I've left hashing to after we've sent the file up to Azure Blob Storage for this reason.  The alternative is we abandon the concept has hashing and storing hashes and ensuring integrality.  Which would make things simpler but equally we can't do you other request which is to prevent storage of duplicate files within the datalake.  There are trade offs.

src/controllers/csv-processor.ts
```diff
@@ -161,55 +179,30 @@ export const moveCSVFromBlobStorageToDatalake = async (dataset: Dataset): Promis
...
+        importRecord.id = randomUUID();
+        importRecord.mime_type = filetype;
         try {
-            logger.debug(`Uploading file ${savedDataFile.id} to blob storage`);
-            await blobStorageService.uploadFile(`${savedDataFile.id}.csv`, buff);
-            return {
-                success: true,
-                dataset: dto
-            };
+            await blobStorageService.uploadFile(`${importRecord.id}.csv`, fileStream);
```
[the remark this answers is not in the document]
j-maynard: We need certainty the operation has completed successfully.  We can't fire and forget because if something goes wrong we could end up loosing the upload and having no data in the cube despite everything that the database tells us.

src/entity2/user.ts
andyjpb: I'm suggesting to just ignore these attributes from OIDC because they're known to be unhelpful.

src/controllers/csv-processor.ts
```diff
@@ -161,55 +179,30 @@ export const moveCSVFromBlobStorageToDatalake = async (dataset: Dataset): Promis
...
+            const resolvedHash = await hashReadableStream(fileStream)
```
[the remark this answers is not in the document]
j-maynard: So if the blobStorage request fails it throws an error... This sits in a try catch box for this reason... If it errors we never make it to the point of generating the hash.  We can only generate the has if we have successfully uploaded the file to Blob Storage.

src/route/dataset-route.ts
andyjpb: It's a UUID: we can (and should) assign it ourselves. That's the point of UUIDs.

src/entity2/import.ts
j-maynard: This might be an error on my part... Let me check

src/controllers/csv-processor.ts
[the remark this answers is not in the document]
andyjpb: See how this is done in spider.py / ugc.py.
We never force materialisation of the stream. We batch it into the stream and peek at it along the way, maintaining small, fixed sized state in the digest object.
So yes, we have to read it, but there's no need to ever materialise the whole stream.

src/controllers/csv-processor.ts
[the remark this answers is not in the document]
andyjpb: We don't need to force everything to be synchronous tho': we only need to know at the point we commit the other objects that rely on it. For example, the DB objects we stash its name in.

src/entity2/revision.ts
j-maynard: Because the ownership is on the part of the Revision not the dataset... So this is the other side of that OneToMany relationship... e.g 1 Dataset, Many revisions.  Here in the revision entity we have only one dataset... but in the dataset entity we have an array of revisions.   Equally that revisions in the dataset are annotted with a `@OneToMany` annotation to match this `@ManyToOne`

src/controllers/csv-processor.ts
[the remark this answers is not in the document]
andyjpb: The problem isn't just when it throws an error, it's when it simply takes longer to process data than the digest function (which should be always as the digest is some simple local compute and the blobstore is some networked IO).

We need to generate the hash *as* we upload it so that we get the benefit of the streaming interface: only having one block in memory at a time.

src/entity2/revision.ts
j-maynard: So I am treating them almost like a linked listed...  The latest revision points to the previous revision and that revision then points to the previous revision, because Revisions can be abandoned you don't get a one to one relationship between revisions so ManyToOne here is correct.  It can also be null as there might not be a previous revision as is the case for the first revision created.

src/entity2/revision.ts
j-maynard: When you create a new revision I rather assumed you'd create a new online cube... the old cubes would in theory still be available or at the very least archived off.

src/entity2/user.ts
j-maynard: I will drop usename in the next commit to this PR.

src/entity2/user.ts
j-maynard: There will be system users I suspect... Right now there's a test user and I am allowing for the possibility to roll a local authentication strategy which is looking increasing like it might be needed.

src/route/dataset-route.ts
j-maynard: That said I might just batch up object creation and save the overall dataset and see what happens in test.

src/route/dataset-route.ts
j-maynard: No... We know if things have been translated based on the number of records attached to each dataset... In their draft form theres always one... English or Welsh.  When we look again to see if the translation step has been completed we are looking for 2 instances of datasetInfo to be present... One in English and one in Welsh:

```
datasetInfo: [
  {
    datasetId: '12345678-1234-1234-1234-123456789012',
    language: 'en-GB',
    title: 'English Title'
  }, {
    datasetId: '12345678-1234-1234-1234-123456789012',
    language: 'cy-GB',
    title: 'Welsh Title'
  }
]
```

We can get everything from the database about what has and has not been completed without using flags to tell us.

src/route/dataset-route.ts
j-maynard: I thought the idea of this was that only active revisions would have an index.  Abandoned revisions wouldn't.  They'd be around and in theory retrievable but they wouldn't have a place in the index.

src/route/dataset-route.ts
andyjpb: That covers the initial state.
But in the steady state there are two translations, both related.
Then a change happens and there are still two translations, one old and one new and we don't know which is which given the metadata.

src/route/dataset-route.ts
j-maynard: I might try batching these up in to a single dataset object creation call and see what happens.

src/route/dataset-route.ts
andyjpb: I think it's just an optimisation for the toposort algorithm so that you don't have to read all revisions and chase all the pointers?
It gives you a correct toposort with a range query and then you can do a single pass to join up the tree.

src/migration/1723729297617-migration.ts
```diff
@@ -0,0 +1,171 @@
...
+                        '',
+                        NULL,
+                        'Test User',
+                        'Test',
+                        'User',  -- Corrected by closing the quote
+                        '',
+                        CURRENT_TIMESTAMP,
+                        CURRENT_TIMESTAMP,
+                        true
+                    );
```
wheelsandcogs, line 154: I would prefer not to add test data in the migrations as we will probably want to run these in prod eventually. We can load test data from the tests before they run, or just manually if it's for manual testing.

wheelsandcogs (commented):
Looks pretty good, I've added quite a few comments but some of it is for my understanding and the rest are bitty things that should be easy to remedy.

If you find I'm too pedantic let me know and I can tweak the levels down a bit 😄

Some general comments:

- We're missing return types on most functions - is this something we want to start getting into the habit of providing? If so, we can add a lint warning when they're missing?
- This is just semantics but I'd say `BlobStorageService` and `CSVProcessor` are both services rather than controllers. Do we want to make `CSVProcessor` a class also?
- The switching of column types based on `environment == test` feels a bit strange, I'm assuming it's to do with some issue you had using the proper types when testing? I've not had to do this before so maybe we can resolve it with test config or mocks?
- There are currently no indexes on any of the entities, we might want to do a pass looking at what fields are used in queries and adding indexes on those fields.

src/entity2/dimension_types.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable no-shadow */
+/* eslint-disable no-unused-vars */
```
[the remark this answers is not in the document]
j-maynard: Yeah... I thought this was odd...  I can't workout if this is a bug in eslint or if I have something configured wrong.  Lets raise an internal GitHub issue on the backend project and we'll revisit it when we have sometime.

src/entity2/users.ts
```diff
@@ -0,0 +1,73 @@
...
+        user.token_expiry = new Date();
+        user.name = 'Test User';
+        user.given_name = 'Test';
+        user.last_name = 'User';
+        user.profile_picture = '';
+        user.created_at = new Date();
+        user.updated_at = new Date();
+        user.active = true;
+        return user;
+    }
```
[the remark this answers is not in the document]
j-maynard: When we wrap up the Auth bits lets move it out then... we need a user entry and way to find it right now.  When we have actual users on the backend we can move this over to the test code.

src/entity2/users.ts
```diff
@@ -0,0 +1,73 @@
...
+    oidc_issuer: string;
+
+    @Column({ type: 'text', nullable: true })
+    access_token: string;
+
+    @Column({ type: 'text', nullable: true })
+    refresh_token: string;
+
+    @Column({ type: 'text', nullable: true })
+    id_token: string;
```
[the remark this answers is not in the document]
j-maynard: So the ID token is supplied by the OIDC provider and is used as part of the conversation between the Identity Provider (e.g. Google, OneLogin, etc) and the backend server.  I was largely under the impression we need to keep this... either in session storage (redis) or the associated with the user in the database.  But we can clear some of this if I address your above comment about clearing up the user class.

src/migration/1723729297617-migration.ts
```diff
@@ -0,0 +1,171 @@
...
+                    DROP TABLE csv_info;
+                    DROP TABLE dimension;
+                    DROP TABLE revision;
+                    DROP TABLE dimension_info;
+                    DROP TABLE dataset_info;
+                    DROP TABLE dataset;
+                    DROP TABLE users;
+                `);
+    }
+}
```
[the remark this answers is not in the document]
j-maynard: No... Largely crafted by hand to ensure it works

src/migration/1723729297617-migration.ts
j-maynard: Again chicken and egg problem...We need a test user we can put things against while we workout the rest of the user journey.  I am happy to remove this once that work is in to place and we ensure we have a way to provisions some initial users.

src/controllers/csv-processor.ts
```diff
@@ -102,201 +117,144 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
-    const datafile: Datafile | undefined = datafiles
-        .filter((filterfile: Datafile) => filterfile.draft === false)
-        .sort(
-            (first: Datafile, second: Datafile) =>
-                new Date(second.creationDate).getTime() - new Date(first.creationDate).getTime()
-        )
-        .shift();
-    const dto = await datasetToDatasetDTO(dataset);
-    if (datafile) {
+    if (fileStream) {
```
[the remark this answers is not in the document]
j-maynard: switched this around

src/controllers/csv-processor.ts
```diff
@@ -102,201 +117,144 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
+        const importRecord = new Import();
+        importRecord.id = randomUUID();
+        importRecord.mime_type = filetype;
+        if (filetype === 'text/csv') {
+            importRecord.filename = `${importRecord.id}.csv`;
+        } else {
+            importRecord.filename = `${importRecord.id}.zip`;
+        }
```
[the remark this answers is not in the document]
j-maynard: Accepted this and put it in to the code

src/controllers/csv-processor.ts
```diff
@@ -102,201 +117,144 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
-                dataset: dto
-            };
+            const promisedHash = hashReadableStream(fileStream)
+                .then((hash) => {
+                    return hash.toString();
+                })
+                .catch((error) => {
+                    throw new Error(`Error hashing stream: ${error}`);
+                });
+            await blobStorageService.uploadFile(`${importRecord.id}.csv`, fileStream);
```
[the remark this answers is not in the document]
j-maynard: That was missed on my part... I've now updated the code to use the filename in full.

src/controllers/csv-processor.ts
```diff
@@ -102,201 +117,144 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
+};
+
+function setupPagination(page: number, total_pages: number): Array<string | number> {
+    const pages = [];
+    if (page !== 1) pages.push('previous');
+    if (page - 1 > 0) pages.push(page - 1);
+    pages.push(page);
+    if (page + 1 <= total_pages) pages.push(page + 1);
+    if (page < total_pages) pages.push('next');
+    return pages;
```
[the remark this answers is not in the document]
j-maynard: This builds the page numbers, and next and previous buttons for use in the frontend.

src/controllers/csv-processor.ts
```diff
@@ -102,201 +117,144 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
+    page: number,
+    size: number,
+    dataset: Dataset,
+    importObj: Import
+): Promise<ViewDTO | ViewErrDTO> {
+    const dataArray: Array<Array<string>> = (await parse(buffer, {
+        delimiter: ','
+    }).toArray()) as string[][];
+    const csvheaders = dataArray.shift();
+    const total_pages = Math.ceil(dataArray.length / size);
```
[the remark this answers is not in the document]
j-maynard: Switched it to camelCase for this... yes we'll tighten up case use in the project.

src/dtos2/filelist.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export interface FileDescription {
+    dataset_id: string;
+    titles: DatasetTitle[];
+}
+
+export interface FileList {
+    files: FileDescription[];
+}
```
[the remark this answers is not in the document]
j-maynard: We can do that... Lets refactor after this lands

src/entity2/dataset.ts
```diff
@@ -0,0 +1,43 @@
+import { Entity, PrimaryGeneratedColumn, Column, BaseEntity, ManyToOne, OneToMany, JoinColumn } from 'typeorm';
+
+import { Users } from './users';
+// eslint-disable-next-line import/no-cycle
+import { Revision } from './revision';
+// eslint-disable-next-line import/no-cycle
+import { DatasetInfo } from './dataset_info';
+// eslint-disable-next-line import/no-cycle
+import { Dimension } from './dimension';
```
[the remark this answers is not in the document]
j-maynard: That would be helpful... I've not found a way around it especially around the entities.  Could be I'm just asking Google, Co-Pilot and ChatGPT the wrong question.  I'd prefer to find a solution but agree, lets land this and we can revisit... I can raise an issue for us to track it as tech debt.

src/dtos2/filelist.ts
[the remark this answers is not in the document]
wheelsandcogs: 👍

src/entity2/dataset.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Entity()
+export class Dataset extends BaseEntity {
+    @PrimaryGeneratedColumn('uuid')
+    id: string;
+
+    @Column({ type: process.env.NODE_ENV === 'test' ? 'datetime' : 'timestamp', default: () => 'CURRENT_TIMESTAMP' })
```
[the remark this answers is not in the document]
j-maynard: I've replaced all the created_at columns with this decorator.

src/entity2/import.ts
```diff
@@ -0,0 +1,58 @@
...
+    @Column({ type: 'varchar', length: 255 })
+    mime_type: string;
+
+    @Column({ type: 'varchar', length: 255 })
+    filename: string;
+
+    @Column({ type: 'varchar', length: 255 })
+    hash: string;
+
+    @Column({ type: process.env.NODE_ENV === 'test' ? 'datetime' : 'timestamp', default: () => 'CURRENT_TIMESTAMP' })
```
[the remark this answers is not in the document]
j-maynard: Changed this as well

src/entity2/dimension_types.ts
```diff
@@ -0,0 +1,11 @@
...
+export enum DimensionType {
```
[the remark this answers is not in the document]
j-maynard: Renamed the file

src/entity2/users.ts
```diff
@@ -0,0 +1,73 @@
+import { Entity, PrimaryGeneratedColumn, Column, BaseEntity } from 'typeorm';
+
+@Entity()
+export class Users extends BaseEntity {
```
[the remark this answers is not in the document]
j-maynard: I've changed this to User (which is how I originally had it until Postgres kept throwing errors about it).  I've decorated the entity with the name "users" to fix Postgres being unhappy.

src/entity2/csv_info.ts
```diff
@@ -0,0 +1,26 @@
+import { Entity, PrimaryColumn, Column, BaseEntity, ManyToOne, JoinColumn } from 'typeorm';
+
+// eslint-disable-next-line import/no-cycle
+import { Import } from './import';
+
+@Entity()
+export class CsvInfo extends BaseEntity {
+    @PrimaryColumn({ name: 'import_id', type: process.env.NODE_ENV === 'test' ? 'text' : 'uuid' })
```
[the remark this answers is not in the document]
j-maynard: If we accept your suggestion of doing all testing against Postgres then we can clean this up.

src/entity2/dimension.ts
```diff
@@ -0,0 +1,60 @@
...
+    @Column({
+        type: process.env.NODE_ENV === 'test' ? 'text' : 'enum',
+        enum: [
+            DimensionType.RAW,
+            DimensionType.TEXT,
+            DimensionType.NUMERIC,
+            DimensionType.SYMBOL,
+            DimensionType.LOOKUP_TABLE,
+            DimensionType.TIME_PERIOD,
+            DimensionType.TIME_POINT
```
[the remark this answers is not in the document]
j-maynard: I've changed this with no ill effects

src/data-source.ts
```diff
@@ -18,7 +18,7 @@ export const datasourceOptions: DataSourceOptions = {
     ssl: true,
     synchronize: false,
     logging: false,
-    entities: [`${__dirname}/entity/*.ts`],
+    entities: [`${__dirname}/entity2/*.ts`],
```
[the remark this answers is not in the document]
j-maynard: Refactor done... entity2 -> entities, dtos2 -> dtos

src/entity2/dataset.ts
```diff
@@ -0,0 +1,43 @@
...
+    @Column({ type: process.env.NODE_ENV === 'test' ? 'datetime' : 'timestamp', nullable: true })
+    live: Date;
+
+    @Column({ type: process.env.NODE_ENV === 'test' ? 'datetime' : 'timestamp', nullable: true })
+    archive: Date;
+
+    @OneToMany(() => DatasetInfo, (datasetInfo) => datasetInfo.dataset, {
+        cascade: true
+    })
+    datasetInfo: Promise<DatasetInfo[]>;
```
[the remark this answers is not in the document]
j-maynard: Refactored all these out

src/entity2/revision.ts
```diff
@@ -0,0 +1,70 @@
...
+    @JoinColumn({ name: 'previous_revision_id' })
+    previous_revision: Promise<RevisionInterface>;
+
+    @Column({ type: 'varchar', length: 255, nullable: true })
+    online_cube_filename: string;
+
+    @Column({ type: process.env.NODE_ENV === 'test' ? 'datetime' : 'timestamp', nullable: true })
+    publish_date: Date;
+
+    @Column({ type: process.env.NODE_ENV === 'test' ? 'datetime' : 'timestamp', nullable: true })
```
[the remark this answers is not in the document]
j-maynard: Switched to using timestamptz in Postgres.

wheelsandcogs (approved):
ship it 🚢

PR #10, opened 2024-08-20
A couple of local dev improvements

wheelsandcogs:
There were a few things missing in the readme with regards to env vars.

Some of the npm scripts were calling things several times (e.g. build or rimraf).

Added a flag to allow SSL to be disabled on the db connection for local dev (default is on if no env var set)


j-maynard (approved):
LGTM 🚀

PR #11, opened 2024-08-21
Use local postgres instance for dev by default

wheelsandcogs:
Adds a compose file for running a local Postgres container which will start by default when running 

```
npm run dev
```

Splits out the code checks so we don't have to run everything on every code change. The old behaviour is still available via:

```
npm run dev:check
```


wheelsandcogs:
Not sure why this hasn't triggered the CI actions?

j-maynard (approved):
LGTM 🚀

PR #12, opened 2024-08-30
SW-183: move auth to backend and use jwt for protected routes

wheelsandcogs:
This adds new routes and config to the backend that provide authentication via Google, OneLogin, and JWT tokens.

The Google and OneLogin providers are currently disabled for NODE_ENV=test. We can revisit this later if/when we have some kind of browser testing (playwright?).

I've done a bit of refactoring (sorry about that) to add a bit more (hopefully) logical structure.


.gitignore
```diff
@@ -37,3 +37,6 @@ yarn.lock
 
 # Ignore local db volume
 .docker/postgres/data
+
+# Ignore certs
+certs/*
```
wheelsandcogs, line 42: This was part of an abandoned attempt to use certs for jwt signing, we can leave in or I can remove?

src/app.ts
```diff
@@ -1,70 +1,45 @@
...
-        fallbackLng: 'en-GB',
-        preload: ['en-GB', 'cy-GB'],
-        debug: false
-    });
-
-export const i18n = i18next;
-export const t = i18next.t;
-export const ENGLISH = 'en-GB';
-export const WELSH = 'cy-GB';
-
```
wheelsandcogs, line 53: I've just moved this to `middleware/translation.ts`

src/controllers/blob-storage.ts
```diff
@@ -7,14 +7,10 @@ import {
...
 import * as dotenv from 'dotenv';
-import pino from 'pino';
 
-dotenv.config();
+import { logger } from '../utils/logger';
 
-export const logger = pino({
-    name: 'StatsWales-Alpha-App: BlobStorageService',
-    level: 'debug'
-});
```
wheelsandcogs, line 17: I've just realised I might've removed some clarity here by reusing the configured logger - [Child loggers](https://github.com/pinojs/pino/blob/main/docs/child-loggers.md) might be a solution here.

src/controllers/auth.ts
```diff
@@ -0,0 +1,81 @@
...
+
+            const payload = { user: sanitiseUser(user) };
+            const expiresIn = process.env.JWT_EXPIRES_IN || DEFAULT_TOKEN_EXPIRY;
+            const token = jwt.sign(payload, process.env.JWT_SECRET || '', { expiresIn });
+
+            res.cookie('jwt', token, {
+                secure: process.env.NODE_ENV !== 'dev',
+                httpOnly: true,
+                domain
+            });
```
wheelsandcogs, line 76: There's a fair bit of duplication here that we can probably clean up, but for now I just wanted to get it working.

src/dtos/dataset-dto.ts
```diff
@@ -350,7 +350,7 @@ export class DatasetDTO {
     static async fromDatasetWithShallowDimensionsAndRevisions(dataset: Dataset): Promise<DatasetDTO> {
         const dto = new DatasetDTO();
         dto.id = dataset.id;
-        dto.creation_date = dataset.creation_date.toISOString();
+        dto.creation_date = dataset.creationDate.toISOString();
```
wheelsandcogs, line 353: If you're ok with it I'll probably do another pass to rename any date columns with `xxx_at` e.g. 
 * created_at
 * updated_at
 * publish_at
 * approved_at
etc

src/entities/csv-info.ts
```diff
@@ -21,6 +21,6 @@ export class CsvInfo extends BaseEntity {
         onDelete: 'CASCADE',
         orphanedRowAction: 'delete'
     })
-    @JoinColumn({ name: 'import_id' })
+    @JoinColumn({ name: 'import_id', foreignKeyConstraintName: 'FK_csv_info_import_id' })
```
wheelsandcogs, line 24: I've added explicit PK and FK naming because the auto-generated names from typeorm are basically indecipherable once they're created. This makes it a pain to remove or modify them later.

src/entities/user.ts
```diff
@@ -1,50 +1,38 @@
...
-        const user = new User();
-        user.id = '12345678-1234-1234-1234-123456789012';
-        user.email = 'test@test.com';
-        user.name = 'Test User';
-        user.givenName = 'Test';
-        user.lastName = 'User';
-        user.createdAt = new Date();
-        user.updatedAt = new Date();
-        user.active = true;
-        return user;
```
wheelsandcogs, line 48: Moved this to `test/helpers/get-user.ts`

src/entities/user.ts
```diff
@@ -1,50 +1,38 @@
...
+    get name(): string {
+        return `${this.givenName} ${this.familyName}`;
```
wheelsandcogs, line 36: We may actually want a `displayName` column at some point, but for now this'll do.

src/entities/user.ts
```diff
@@ -1,50 +1,38 @@
...
 
-    @Column({ nullable: true })
-    givenName: string;
+    @Column({ name: 'given_name', nullable: true })
+    givenName?: string;
 
-    @Column({ nullable: true })
-    lastName: string;
+    @Column({ name: 'family_name', nullable: true })
+    familyName?: string;
```
wheelsandcogs, line 27: I think `given_name` and `family_name` are in the openid connect spec, so we might as well use those keys.

src/enums/data-location.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable no-shadow */
+/* eslint-disable no-unused-vars */
+export enum DataLocation {
+    BLOB_STORAGE = 'BlobStorage',
+    DATA_LAKE = 'Datalake'
+}
```
wheelsandcogs, line 6: Moved some of the enums to their own files - still need to investigate what the deal is with eslint hating on them.

src/middleware/passport-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+    passport.use(
+        'jwt',
+        new JWTStrategy(
+            {
+                jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+                secretOrKey: process.env.JWT_SECRET!
+            },
+            async (jwtPayload, done) => {
+                logger.info('request authenticated with JWT');
+                const user = await userRepository.findOneBy({ id: jwtPayload?.user?.id });
```
wheelsandcogs, line 26: I decided to always reload the user from the db, as I can't think of a good reason not to.

src/middleware/passport-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+    return createPrivateKey({ key: Buffer.from(privateKey, 'base64'), type: 'pkcs8', format: 'der' });
+};
+
+export const initPassport = async (userRepository: Repository<User>): Promise<void> => {
+    passport.use(
+        'jwt',
+        new JWTStrategy(
+            {
+                jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+                secretOrKey: process.env.JWT_SECRET!
```
wheelsandcogs, line 22: This can be a cert, but for now a secret will do.

src/middleware/passport-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+            },
+            async (jwtPayload, done) => {
+                logger.info('request authenticated with JWT');
+                const user = await userRepository.findOneBy({ id: jwtPayload?.user?.id });
+                return done(null, user);
+            }
+        )
+    );
+
+    if (process.env.AUTH_PROVIDERS?.includes('onelogin')) {
```
wheelsandcogs, line 32: We can configure which auth providers get enabled via env config. This is useful for turning them off in CI, or if we only want to allow Google in dev for example.

src/middleware/passport-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+                        ui_locales: 'en',
+                        scope: 'openid email'
+                        // TODO: we need to update our OneLogin config to be able to request the below claims
+                        // vtr: ['P2.Cl.Cm'],
+                        // claims: {
+                        //     userinfo: {
+                        //         'https://vocab.account.gov.uk/v1/coreIdentityJWT': null
+                        //     }
+                        // }
+                    }
```
wheelsandcogs, line 64: Getting a working config for this is what took days of head scratching 😬 

As mentioned in the TODO, we need to bump some settings before we can fetch given and family name.

src/middleware/passport-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+                },
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                async (tokenset: TokenSet, userInfo: UserinfoResponse, cb: any): Promise<void> => {
+                    logger.debug('auth callback from onelogin received');
+
+                    if (userInfo.email === undefined) {
+                        logger.error('onelogin auth failed: account has no email address');
+                        cb(null, undefined, {
+                            message: 'onelogin account does not have an email, use another provider'
```
wheelsandcogs, line 73: This shouldn't ever happen for OneLogin as far as I understand it, but might as well check.

src/middleware/passport-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+                    }
+
+                    try {
+                        logger.debug('checking if user has previously logged in...');
+                        const existingUser = await userRepository.findOneBy({ email: userInfo.email });
+
+                        if (existingUser && existingUser.provider !== 'onelogin') {
+                            logger.error('onelogin auth failed: email was registered via another provider');
+                            cb(null, undefined, { message: 'User is already registered via another provider' });
+                            return;
```
wheelsandcogs, line 85: We could potentially support multiple oauth providers per account, but that would require a more complex user setup so I've ignored it for now.

src/middleware/passport-auth.ts
```diff
@@ -0,0 +1,168 @@
...
+                    callbackURL: `${process.env.BACKEND_URL}/auth/google/callback`,
+                    scope: ['openid', 'profile', 'email']
+                },
+                async (accessToken, refreshToken, profile, cb): Promise<void> => {
+                    logger.debug('auth callback from google received');
+
+                    if (!profile?._json?.email) {
+                        logger.error('google auth failed: account has no email address');
+                        cb(null, undefined, {
+                            message: 'google account does not have an email, use another provider'
```
wheelsandcogs, line 127: For google this can actually happen if the user created an account with just a phone number (not sure if this is still a thing, but the docs say to check).

src/middleware/rate-limiter.ts
```diff
@@ -0,0 +1,13 @@
...
+    windowMs: parseInt(process.env.RATE_LIMIT_WINDOW_MS || '60000', 10),
+    max: parseInt(process.env.RATE_LIMIT_MAX_REQ || '100', 10),
+    standardHeaders: true,
+    legacyHeaders: false,
+    handler: (req, res) => {
+        res.status(429).json({
+            message: 'Too many requests, please try again later.'
+        });
+    }
+});
```
wheelsandcogs, line 13: GitHub was complaining about no rate limiter on the API, so I copied the middleware from the frontend.

src/migrations/1725962477957-initial-schema.ts
```diff
@@ -0,0 +1,127 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
```
wheelsandcogs, line 2: This file was 100% generated by `npm run migration:generate` and then just auto cleaned up with prettier and eslint.

Happy to add whitespace to make it more human-readable if you prefer?

src/controllers/auth.ts
j-maynard: There is... But it wasn't that which caught my eye.   I was wondering if we might not want these in seperate files `google-auth.ts` and `onelogingov-auth.ts` in case we end up adding other auth providers like ActiceDirectory in the future.  Its not something we need to do now just something for us to consider.

src/app.ts
j-maynard: This is good and something I should have done a while ago.

.gitignore
j-maynard: Leave it for this PR... we can remove it later if we decide for sure we don't want to try to use certfiicates

package.json
```diff
@@ -12,30 +12,38 @@
...
+    "test": "NODE_ENV=test jest --coverage",
+    "test:ci": "NODE_ENV=test jest --ci --coverage --config=jest.config.ts",
     "check": "npm-run-all prettier:fix lint:fix test build",
-    "predev": "docker compose up -d",
+    "predev": "docker compose up -d db-dev",
     "dev:check": "npm-run-all check dev",
-    "dev": "nodemon --watch src -e ts,ejs --exec npm run start",
-    "start": "npm run build && node dist/server.js",
+    "dev": "NODE_ENV=dev nodemon --watch src -e ts,ejs --exec npm run start",
+    "start": "npm run build && node dist/server.js | pino-colada",
```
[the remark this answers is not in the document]
j-maynard: You can leave this if it makes your life easier then we can keep it as that will keep things nice and clean

src/controllers/auth.ts
wheelsandcogs: Makes sense 👍

Next time we touch auth I'll split them out.

src/controllers/datalake.ts
```diff
@@ -1,16 +1,9 @@
 import { basename } from 'path';
 import { Readable } from 'stream';
 
-import * as dotenv from 'dotenv';
 import { DataLakeServiceClient, StorageSharedKeyCredential } from '@azure/storage-file-datalake';
-import pino from 'pino';
 
-dotenv.config();
-
-export const logger = pino({
```
j-maynard, line 10: We'll need to add a child logger here.  I'll raise an issue for this so that we can go back to it rather than blocking this PR from moving forward

src/db/data-source.ts
```diff
@@ -0,0 +1,37 @@
...
+    password: DB_PASSWORD,
+    database: DB_DATABASE,
+    ssl: DB_SSL !== 'false' ?? true,
+    synchronize: false,
+    logging: false,
+    entities: [`${__dirname}/../entities/*.{ts,js}`],
+    migrations: [`${__dirname}/../migrations/*.{ts,js}`]
+};
+
+const { TEST_DB_HOST, TEST_DB_PORT, TEST_DB_USERNAME, TEST_DB_PASSWORD, TEST_DB_DATABASE } = process.env;
```
j-maynard, line 21: I can't help feeling this violates the rule on not poluting our code with test code when we can load in a different configuration when doing the tests.

src/entities/user.ts
j-maynard: E-mail, Name, Given Name and Family Name are all part of the OIDC profile spec and should be populated from the profile ideally.  Name isn't always a conjunction of Given Name and Family Name for exmaple

src/entities/user.ts
j-maynard: See comment about `name` above...  Its part of the OIDC Profile and we should probably be capturing it.  I would be surprised if Welsh Gov users aren't just `first name` `family name` but its probably a bad asumption to make as someone out there might surprise us.

src/enums/data-location.ts
j-maynard: So the answer as to why eslint dislike enums is that they're not native to JavaScript and when the transpiler converts the code it does some screwy stuff to make enums work.

j-maynard, line 6: Could you add
```suggestion
    UNKNOWN = 'Unknown'
```
Turns out there is a need for an unknown location especially when testing things

src/middleware/passport-auth.ts
j-maynard: I agree... A secret will do for now.  We may want to look at certs again when we have more time.

src/middleware/passport-auth.ts
j-maynard: Could bury the user in a session and get it from there but honestly it wouldn't save any overhead or time.  So this is fine.  If a user is disabled it means we can lock out a user more quickly rather than waiting for their session to expire.

src/controllers/blob-storage.ts
wheelsandcogs: done 👍

src/controllers/datalake.ts
wheelsandcogs: done 👍

src/db/data-source.ts
wheelsandcogs: Yeah, there's probably a better way of doing this, will revisit in another pass. Just needed something that would get postgres working in local/ci tests for now.

src/middleware/passport-auth.ts
j-maynard: Maybe raise an issue against the TODO here on GitHub so we have some sight of it away from the code and we can come back look again when time or need allows.

src/entities/user.ts
wheelsandcogs: Yeah this already happens for google, but we need higher perms for OneLogin to request  `https://vocab.account.gov.uk/v1/coreIdentityJWT` claim which includes names.

For now we've got enough to create an account.

No display name returned by OneLogin as far as I can see, so we might need to fudge one for now, hence the getter.

src/middleware/passport-auth.ts
j-maynard: We need to... Maybe someones impersonating OneLogin maybe something has gone wrong.  We should always check fot the unlikely especially when we have no control over it.

src/entities/user.ts
wheelsandcogs: as above, we don't have access to names from OneLogin yet so this is just a placeholder :)

src/enums/data-location.ts
wheelsandcogs: Yeah but I've used enums in typescript tons before without eslint crying - I think we just need to fiddle with the eslint config.

wheelsandcogs: done 👍

src/middleware/passport-auth.ts
wheelsandcogs: So once you're authed, `req.user` is set with the user object from here and available in any authed route.

src/middleware/passport-auth.ts
j-maynard: I'm cool with doing what the Docs say...  I don't think its a thing any more and given we're limiting to those with Marvell Google logins definately won't be the case.

src/middleware/rate-limiter.ts
j-maynard: Seems sensible to me... We do need to keep the code sniffer happy generally.

src/migrations/1725962477957-initial-schema.ts
j-maynard: This should be fine.  I don't imagine us looking to closely at it as its generated by TypeOrm unless something goes wrong.

src/route/healthcheck.ts
```diff
@@ -1,25 +1,31 @@
...
-import { logger } from '../app';
-import { DataLakeService } from '../controllers/datalake';
+// import { logger } from '../utils/logger';
+// import { DataLakeService } from '../controllers/datalake';
+import { sanitiseUser } from '../utils/sanitise-user';
+import { User } from '../entities/user';
 
-const dataLakeService = new DataLakeService();
-export const healthcheck = Router();
+const healthcheck = Router();
```
j-maynard, line 10: We do need a complete `/` healthcheck to pass to Docker and the Container service to tell us when things aren't working.  Really this should include checking we have the following:
- Database connection
- Datalake connection
- BlobStorage connection

As these are required for the service to work and if they stop working for any reason its good to be able to look at the health check to see thats the case.

src/entities/csv-info.ts
j-maynard: Happy with this and letting TypeOrm do the heavy lifting on migration generation.

src/middleware/passport-auth.ts
j-maynard: This is good... I like configuration to turn things off and on.  We might in production want or need to remove Marvell Google for exmaple or add social logins in the future so its good to be able to turn these things off and on.  Can you add this to the README?

src/middleware/passport-auth.ts
wheelsandcogs: done

src/middleware/passport-auth.ts
wheelsandcogs: done - SW-190

src/route/healthcheck.ts
wheelsandcogs: I agree, we should have a healthcheck/<service> for any connected services that we can use for monitoring.

I had to disable the datalake one temporarily because it is failing in CI when attempting to connect to `your-storage-account-name.x.y.z` endpoint to fetch creds.

PR #13, opened 2024-09-09
SW-138 Update backend for sources and dimensions

j-maynard:
This updates the back with 2 new routes.

The is for confirming the uploaded fact table and moving it from tempoary blob storage to tghe datalate.

The second route handles labelling sources as Dimension, Footnotes and DataValues as well as the option to ignore a source entirly.

Also include is some bug fixes within the Data model and database script.


.eslintrc
```diff
@@ -24,7 +24,7 @@
     "no-process-env": 0,
     "no-inline-comments": 0,
     "line-comment-position": 0,
-    "no-warning-comments": 1
+    "no-warning-comments": 0
```
wheelsandcogs, line 27: This is `TODO` comments right? Not a fan?

j-maynard: Am a Fan... Right now need... hmm.. I wonder why I set that 🤔 
Let me fix it when you're done with the review.

src/route/dataset.ts
```diff
@@ -219,37 +233,43 @@ router.get('/', async (req: Request, res: Response) => {
...
     res.json(dto);
 });
 
+// DELETE /api/dataset/:dataset_id
+// Returns a shallow dto of the dataset with the given ID
+// Shallow gives the revisions and dimensions of the dataset only
+router.delete('/:dataset_id', async (req: Request, res: Response) => {
+    const datasetID: string = req.params.dataset_id.toLowerCase();
+    const dataset = await validateDataset(datasetID, res);
+    if (!dataset) return;
```
wheelsandcogs, line 249: Do we want a 404 here, or is it handled by a fall-through route?

src/route/dataset.ts
```diff
@@ -299,12 +319,66 @@ router.get('/:dataset_id/revision/by-id/:revision_id', async (req: Request, res:
...
+// POST /api/dataset/:dataset_id/revision/id/:revision_id/import
+// Creates a new import on a revision.  This typically only occurs when a user
+// decides the file they uploaded wasn't correct.
+router.post(
+    '/:dataset_id/revision/by-id/:revision_id/import',
+    upload.single('csv'),
+    async (req: Request, res: Response) => {
+        const datasetID: string = req.params.dataset_id.toLowerCase();
+        const dataset = await validateDataset(datasetID, res);
+        if (!dataset) return;
```
wheelsandcogs, line 331: 404?

src/route/dataset.ts
```diff
@@ -299,12 +319,66 @@ router.get('/:dataset_id/revision/by-id/:revision_id', async (req: Request, res:
...
+        res.json(uploadDTO);
+    }
+);
+
+// GET /api/dataset/:dataset_id/revision/id/:revision_id/import/by-id/:import_id
+// Returns details of an import with its sources
+router.get('/:dataset_id/revision/by-id/:revision_id/import/by-id/:import_id', async (req: Request, res: Response) => {
+    const datasetID: string = req.params.dataset_id.toLowerCase();
+    const dataset = await validateDataset(datasetID, res);
+    if (!dataset) return;
```
wheelsandcogs, line 365: 404?

src/route/dataset.ts
```diff
@@ -299,12 +319,66 @@ router.get('/:dataset_id/revision/by-id/:revision_id', async (req: Request, res:
     res.json(dto);
 });
 
+// POST /api/dataset/:dataset_id/revision/id/:revision_id/import
+// Creates a new import on a revision.  This typically only occurs when a user
+// decides the file they uploaded wasn't correct.
+router.post(
+    '/:dataset_id/revision/by-id/:revision_id/import',
```
wheelsandcogs, line 326: Do we want/need the `by-id` part of the path?

src/route/dataset.ts
```diff
@@ -379,3 +454,124 @@ router.get(
...
+// PATCH /api/dataset/:dataset_id/revision/id/:revision_id/import/id/:import_id/confirm
+// Moves the file from temporary blob storage to datalake and creates sources
+// returns a JSON object with the current state of the revision including the import
+// and sources created from the import.
+router.patch(
+    '/:dataset_id/revision/by-id/:revision_id/import/by-id/:import_id/confirm',
+    async (req: Request, res: Response) => {
+        const datasetID: string = req.params.dataset_id.toLowerCase();
+        const dataset = await validateDataset(datasetID, res);
+        if (!dataset) return;
```
wheelsandcogs, line 467: 404?

src/route/dataset.ts
```diff
@@ -379,3 +454,124 @@ router.get(
...
+
+// DELETE /:dataset_id/revision/by-id/:revision_id/import/by-id/:import_id
+// Removes the import record and associated file from BlobStorage clearing the way
+// for the user to upload a new file for the dataset.
+router.delete(
+    '/:dataset_id/revision/by-id/:revision_id/import/by-id/:import_id',
+    async (req: Request, res: Response) => {
+        const datasetID: string = req.params.dataset_id.toLowerCase();
+        const dataset = await validateDataset(datasetID, res);
+        if (!dataset) return;
```
wheelsandcogs, line 510: 404?

src/controllers/csv-processor.ts
```diff
@@ -120,14 +125,14 @@ function validateParams(page_number: number, max_page_number: number, page_size:
...
+export const uploadCSVToBlobStorage = async (fileStream: Readable, filetype: string): Promise<FileImport> => {
     const blobStorageService = new BlobStorageService();
     if (!fileStream) {
         logger.error('No buffer to upload to blob storage');
         throw new Error('No buffer to upload to blob storage');
     }
-    const importRecord = new Import();
-    importRecord.id = randomUUID();
+    const importRecord = new FileImport();
+    importRecord.id = randomUUID().toLowerCase();
```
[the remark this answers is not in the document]
j-maynard: In this case it was causing an issue... which is why I assigned the IDs here.

src/controllers/csv-processor.ts
```diff
@@ -316,18 +355,82 @@ export const processCSVFromBlobStorage = async (
...
+        source.csvField = header.name;
+        source.action = SourceAction.UNKNOWN;
+        source.revision = Promise.resolve(revision);
+        source.import = Promise.resolve(importObj);
+        return source;
+    });
+    const freshFileImport = await FileImport.findOneBy({ id: importObj.id });
+    if (!freshFileImport) {
+        throw new Error('Import not found');
+    }
```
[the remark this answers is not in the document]
j-maynard: yes... Otherwise ESLint and TypeScript complain that the object could be undefined

src/controllers/datalake.ts
```diff
@@ -37,6 +37,50 @@ export class DataLakeService {
...
+    //     const fileSystemClient = this.serviceClient.getFileSystemClient(fileSystemName);
+    //     const directoryClient = fileSystemClient.getDirectoryClient(defaultDirectoryName);
+    //     const fileClient = directoryClient.getFileClient(fileName);
+    //     await fileClient.create();
+    //     await fileClient.uploadStream(fileContent, {
+    //         chunkSize: 8 * 1024 * 1024,
+    //         maxConcurrency: 5
+    //     });
+    //     await fileClient.flush(fileContent.readableLength);
+    // }
```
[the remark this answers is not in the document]
j-maynard: Do we ever want to move to using Streaming in the Future?  This was part of that work, abandoned for speed of delivery right now.  I am happy to remove this from here but it'll be back as we have a ticket from Andy about it.

src/controllers/datalake.ts
```diff
@@ -37,6 +37,50 @@ export class DataLakeService {
...
+
+    public async uploadFileStream(fileName: string | undefined, fileContent: Readable) {
+        if (fileName === undefined) {
+            throw new Error('File name is undefined');
```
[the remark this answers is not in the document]
j-maynard: I can drop the undefined from the top. and remove the check, this is based on older code where there was a chance the filename might have been undefined.

src/controllers/datalake.ts
```diff
@@ -37,6 +37,50 @@ export class DataLakeService {
...
+        let position = 0;
+        for await (const chunk of fileContent) {
+            const chunkSize = chunk.length;
+            // Append the chunk at the current position
+            await fileClient.append(chunk, position, chunkSize);
+            position += chunkSize;
+        }
+        // Flush and commit the file
+        await fileClient.flush(position);
+    }
```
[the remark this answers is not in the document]
j-maynard: I think so... I'd need to look at all the code to make sure.  If it is I can remove it as part of this PR for you

src/controllers/csv-processor.ts
```diff
@@ -316,18 +355,82 @@ export const processCSVFromBlobStorage = async (
...
+        const fileStream = await blobStorageService.getReadableStream(importObj.filename);
+        await datalakeService.uploadFileStream(importObj.filename, fileStream);
+        await blobStorageService.deleteFile(importObj.filename);
+    } catch (err) {
+        logger.error(err);
+        throw new Error('Error moving file to datalake');
+    }
+};
+
+export const removeTempfileFromBlobStorage = async (importObj: FileImport) => {
```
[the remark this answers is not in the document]
j-maynard: No... Because the moveFile only handles copying the file from BlobStorage to the Datalake... Not the remove operation.  I kept them separate as they can both fail and its good to have an over view of why it failed.  Technically removing the file probably isn't fatal to the operation of the system.  So we might choose to not return an error if we fail to remove the file.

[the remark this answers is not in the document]
j-maynard: Yes... If the user backs out of the dataset upload... Decides they've uploaded the wrong file this cleans up that file.  See the `router.delete('/:dataset_id/revision/by-id/:revision_id/import/by-id/:import_id')` route which is part of the transition from previewing the datafile to displaying the sources or choosing to upload a different file.

src/controllers/dimension-processor.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export const validateDimensionCreationRequest = async (
+    dimensionCreationDTO: DimensionCreationDTO[]
+): Promise<ValidatedDimensionCreationRequest> => {
+    let datavalues: DimensionCreationDTO | null = null;
+    let footnotes: DimensionCreationDTO | null = null;
+    const dimensions: DimensionCreationDTO[] = [];
+    const ignore: DimensionCreationDTO[] = [];
+    await Promise.all(
+        dimensionCreationDTO.map(async (sourceInfo) => {
```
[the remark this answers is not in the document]
j-maynard: Well according to Andys demo in the Architecture meeting this afternoon... `9` is the current max.  There isn't a Max for the purposes of what has been proposed at this point.

src/controllers/dimension-processor.ts
```diff
@@ -0,0 +1,137 @@
...
+            })
+        );
+        footnoteDimension.dimensionInfo = Promise.resolve(footnoteDimensionInfo);
+        footnoteDimension.type = DimensionType.FOOTNOTE;
+        footnoteDimension.dataset = Promise.resolve(dataset);
+        footnoteDimension.startRevision = Promise.resolve(revision);
+        const source = await Source.findOne({ where: { id: footnotes.sourceId } });
+        if (!source) {
+            throw new Error(`Source with id ${footnotes.sourceId} not found`);
+        }
```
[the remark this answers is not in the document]
j-maynard: again... This resolves the issue in typescript that the result of a findOne can result in an undefined so to prevent a compiler issue you have to check to make sure the thing you requested really is a thing rather than an undefined.

src/controllers/dimension-processor.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    await Promise.all(
+        dimensions.map(async (dimensionCreationDTO: DimensionCreationDTO) => {
+            const dimension = new Dimension();
+            dimension.id = crypto.randomUUID().toLowerCase();
+            dimension.type = DimensionType.RAW;
+            const source = await Source.findOne({ where: { id: dimensionCreationDTO.sourceId } });
+            if (!source) {
+                throw new Error(`Source with id ${dimensionCreationDTO.sourceId} not found`);
+            }
```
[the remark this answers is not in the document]
j-maynard: See above... The compiler complains because the result of the findOne is either a Source or an undefined and you have to prove its not an undefined.

src/dtos/dataset-dto.ts
```diff
@@ -27,6 +27,19 @@ export class SourceDTO {
...
+        const sourceDto = new SourceDTO();
+        sourceDto.id = source.id;
+        sourceDto.import_id = (await source.import).id;
+        sourceDto.revision_id = (await source.revision).id;
+        // sourceDto.lookup_table_revision_id = (await source.lookupTableRevision)?.id;
+        sourceDto.csv_field = source.csvField;
+        sourceDto.action = source.action;
+        sourceDto.type = source.type;
+        return sourceDto;
+    }
```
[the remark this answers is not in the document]
j-maynard: We should clean this up... I'll raise a new PR to separate this out once the current work is done.

src/dtos/dataset-dto.ts
```diff
@@ -142,6 +156,18 @@ export class RevisionDTO {
...
+                impDto.sources = await Promise.all(
+                    (await imp.sources).map(async (source: Source) => {
+                        const sourceDto = new SourceDTO();
+                        sourceDto.id = source.id;
+                        sourceDto.import_id = (await source.import).id;
+                        sourceDto.revision_id = (await source.revision).id;
+                        sourceDto.csv_field = source.csvField;
+                        sourceDto.action = source.action;
+                        sourceDto.type = source.type;
+                        return sourceDto;
```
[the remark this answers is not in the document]
j-maynard: Yes... although I like the control to ensure we're not nesting objects to much.  For example if you tried to return the whole dataset.  You'd ended up with multiples of the dimension objects for example as they are referenced by both revision and dataset.... Multiple sources as well because they are referenced by both import and dimension.  I do like the control of specifying exactly what we're returning from a given DTO to prevent that.

Lets have a look at this as we break down the file in to individual DTO files for each object.

src/middleware/translation.ts
```diff
@@ -23,4 +23,6 @@ i18next
         debug: false
     });
 
-export { i18next, i18nextMiddleware, ENGLISH, WELSH };
+const t = i18next.t;
+
+export { t, i18next, i18nextMiddleware, ENGLISH, WELSH };
```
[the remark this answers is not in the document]
j-maynard: I went with `t` because that's what is documented against the library.  So it was more consistency... t in our context is always translate and a reference to `i18next` library.

src/resources/locales/en-GB.json
```diff
@@ -46,6 +47,12 @@
             "upload_to_datalake": "Error uploading the file to the datalake"
         }
 	},
+    "dimensionInfo": {
+        "footnotes": {
+            "title": "Footnotes",
+            "description": "Footnotes for the data"
+        }
+    },
```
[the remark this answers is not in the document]
j-maynard: So generic stuff at the top, errors and routes at the bottom... Everything else ideally organised by route or journey.  I've started trying to get to this with the frontend file.  Ideally I think there should be one translation matrix shared across both project...  Equally longer term... Maybe its not a file at all but a Database table which is something that `i18next` supports.

src/entities/source.ts
```diff
@@ -46,6 +48,9 @@ export class Source extends BaseEntity {
     @Column({ name: 'csv_field', type: 'text' })
     csvField: string;
 
-    @Column({ type: 'enum', enum: Object.values(SourceAction), nullable: false })
-    action: string;
+    @Column({ type: 'enum', enum: Object.values(SourceAction), nullable: true })
+    action: SourceAction;
```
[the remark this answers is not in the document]
j-maynard: So it should only be null at the point we create it and wait for the user to tell us what it is and what to do with it... At all other time it should be there... Which raises the question should we treat is an optional?

src/entities/source.ts
```diff
@@ -46,6 +48,9 @@ export class Source extends BaseEntity {
...
+
+    @Column({ type: 'enum', enum: Object.values(SourceType), nullable: true })
+    type: SourceType;
```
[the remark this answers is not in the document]
j-maynard: Aagin... It should only be null during the transition point of waiting for the user to tell us what to do with it.

src/route/dataset.ts
j-maynard: So no dataset means we've already set the status code on the response in the `validateDataset` function.  So no dataset we just return because we've done everything we need to... If that makes sense?

src/route/dataset.ts
j-maynard: For revision yes... For sure... You might want to look up a revision by index... Equally you might want to look up a dataset by code or name.  Same applies to dimension as well... for import it just becomes about consistency.

src/controllers/dimension-processor.ts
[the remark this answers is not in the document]
wheelsandcogs: Try replacing `findOne(...)` with `findOneOrFail(...)`

src/controllers/datalake.ts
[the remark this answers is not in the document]
wheelsandcogs: There's a function with the same name below that is not commented out, so I assumed this was an old replaced version of `uploadFileStream()`?

src/route/dataset.ts
wheelsandcogs: 👍

src/dtos/dataset-dto.ts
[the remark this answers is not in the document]
wheelsandcogs: 👍

src/dtos/dataset-dto.ts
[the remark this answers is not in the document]
wheelsandcogs: 👍

src/resources/locales/en-GB.json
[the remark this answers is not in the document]
wheelsandcogs: 👍

src/route/dataset.ts
wheelsandcogs: ah ok I had missed that validateDataset was manipulating the response.

In that case, what about making it a middleware? Something like:

```typescript
const validateDataset = async (req, res, next) => {
  try {
    res.locals.dataset = await Dataset.findOneByOrFail({ id: req.params.dataset_id });
  catch(err) {
    res.status(404);
    return res.json({ message: 'Dataset not found.' });
  }
};
```

and then apply it to the routes that need the dataset:

```typescript
router.delete('/:dataset_id', validateDataset, async (req...) => {
  const dataset = res.locals.dataset as Dataset;
});
```

package.json
```diff
@@ -12,11 +12,13 @@
...
     "build": "npm-run-all clean tsc copy-assets",
-    "pretest": "docker compose up -d db-test",
+    "pretest-docker": "docker compose up -d db-test",
+    "pretest-podman": "podman compose up -d db-test",
     "test": "NODE_ENV=test jest --coverage",
     "test:ci": "NODE_ENV=test jest --ci --coverage --config=jest.config.ts",
     "check": "npm-run-all prettier:fix lint:fix test build",
-    "predev": "docker compose up -d db-dev",
+    "predev-docker": "docker compose up -d db-dev",
+    "predev-podman": "podman compose up -d db-dev",
```
wheelsandcogs, line 21: This is going to break the current script functionality - npm will automatically run `preX` scripts when you `npm run X`, so `npm run test` and `npm run dev` will stop automatically bringing up the relevant containers.

https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts

Maybe a quick alternative here is to add a docker alias for podman in your `.zshrc`?

```
alias docker='podman'
```

That should allow the pre scripts to work as intended.

j-maynard: "Should!"... Actually yeah we can probably do that... The issue was the mounts in the docker-compse tripping up podman.  I'll revert that :-)

src/controllers/dimension-processor.ts
```diff
@@ -0,0 +1,138 @@
...
+import { SourceType } from '../enums/source-type';
+import { i18next, t } from '../middleware/translation';
+import { SourceAction } from '../enums/source-action';
+
+export interface ValidatedDimensionCreationRequest {
+    datavalues: DimensionCreationDTO | null;
+    footnotes: DimensionCreationDTO | null;
+    dimensions: DimensionCreationDTO[];
+    ignore: DimensionCreationDTO[];
+}
```
wheelsandcogs, line 17: We should probably extract this to it's own file.

If `datavalues` and `footnotes` are optional then I'd avoid using `null` and just mark the props with a `?`.

src/dtos/dimension-dto.ts
```diff
@@ -0,0 +1,58 @@
...
+    sources?: SourceDTO[];
+    dimensionInfo?: DimensionInfoDTO[];
+    dataset_id?: string;
+
+    static async fromDimension(dimension: Dimension): Promise<DimensionDTO> {
+        const dimDto = new DimensionDTO();
+        dimDto.id = dimension.id;
+        dimDto.type = dimension.type;
+        dimDto.start_revision_id = (await dimension.startRevision).id;
+        dimDto.finish_revision_id = (await dimension.finishRevision)?.id || '';
```
wheelsandcogs, line 38: If this is optional then you don't need the empty string fallback, undefined is fine.

src/dtos/fileimport-dto.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    static async fromImport(importEntity: FileImport): Promise<ImportDTO> {
+        const dto = new ImportDTO();
+        dto.id = importEntity.id;
+        const revision = await importEntity.revision;
+        dto.revision_id = revision.id;
+        dto.mime_type = importEntity.mimeType;
+        dto.filename = importEntity.filename;
+        dto.hash = importEntity.hash;
+        dto.uploaded_at = importEntity.uploadedAt?.toISOString() || '';
```
wheelsandcogs, line 25: If this is optional then you don't need the empty string fallback, undefined is fine.

src/dtos/revision-dto.ts
```diff
@@ -0,0 +1,55 @@
...
+    static async fromRevision(revision: Revision): Promise<RevisionDTO> {
+        const revDto = new RevisionDTO();
+        revDto.id = revision.id;
+        revDto.revision_index = revision.revisionIndex;
+        revDto.dataset_id = (await revision.dataset).id;
+        revDto.creation_date = revision.creationDate.toISOString();
+        revDto.previous_revision_id = (await revision.previousRevision)?.id || '';
+        revDto.online_cube_filename = revision.onlineCubeFilename;
+        revDto.publish_date = revision.publishDate?.toISOString() || '';
+        revDto.approval_date = revision.approvalDate?.toISOString() || '';
```
wheelsandcogs, line 28: Same as above, don't stick empty strings in if we mean undefined.

src/controllers/datalake.ts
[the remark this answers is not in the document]
j-maynard: Removed the commented code.

src/controllers/datalake.ts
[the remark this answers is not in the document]
j-maynard: Removed the undefined from the top and the checks.

src/controllers/datalake.ts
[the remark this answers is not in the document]
j-maynard: Going to leave this alone for this PR... and remove it in a follow up PR.  I'd like to have the blob storage service and datalake service share the same basic API points.  We can combine this work with the moving everything to streams work which will also need to be done.

src/entities/source.ts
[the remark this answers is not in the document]
j-maynard: Made them optional

src/route/dataset.ts
j-maynard: Let look at maybe doing this as follow up and get this landed.  I am thinking about generally moving towards `findOneByOrFail` for most database calls.

src/controllers/dimension-processor.ts
j-maynard: So its only used here and only used for processing the request from the client internally.  We sort the request in to various parts as each is processed slightly differently and I suspect there might be some evolution here later.  I don't expect this interface to be used outside of the dimension creation route and dimension processor so I think it makes sense to keep it with the thing that chiefly uses it.

j-maynard: As for datavalues and footnotes being optional... I don't think they really are but we haven't had any thing in the specifications which have been given to us where we can insist on these things.  Footnotes might be optional longer term but datavalues are almost certainly going to be required.  This is an area for iteration later I think.

src/dtos/dimension-dto.ts
j-maynard: Fixed this

src/dtos/fileimport-dto.ts
j-maynard: Fixed this.

src/dtos/revision-dto.ts
j-maynard: Fixed this.

package.json
```diff
@@ -12,7 +12,8 @@
     "lint:ci": "eslint . --ext .ts",
     "lint:fix": "eslint . --ext .ts --fix",
     "build": "npm-run-all clean tsc copy-assets",
-    "pretest": "docker compose up -d db-test",
+    "pretest-docker": "docker compose up -d db-test",
+    "pretest-podman": "podman compose up -d db-test",
```
wheelsandcogs, line 16: Can we revert this as well please? Should be same solution (alias docker=podman)

src/entities/dataset_info.ts
```diff
@@ -0,0 +1,42 @@
...
+    dataset: Promise<Dataset>;
+
+    @CreateDateColumn({ name: 'created_at' })
+    createdAt: Date;
+
+    // This column should be the same across all languages
+    // If one is updated and the others aren't then mark as needing translation
+    @UpdateDateColumn({ name: 'updated_at', nullable: true })
+    updatedAt: Date;
+}
```
wheelsandcogs, line 42: Something a bit gnarly here after your rebase - we've got both `dataset-info.ts` and `dataset_info.ts` 

Let's stick to kebab-case file names.

test/helpers/test-data-source.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const datasourceOptions: DataSourceOptions = {
+    name: 'default',
+    type: 'better-sqlite3',
+    database: ':memory:',
+    synchronize: true,
+    logging: false,
+    entities: [Dataset, DatasetInfo, Revision, FileImport, CsvInfo, Source, Dimension, DimensionInfo, User],
+    subscribers: []
+};
```
wheelsandcogs, line 25: This has come back from the dead also?

src/entities/dataset_info.ts
j-maynard: Removed this file

src/entities/dimension_info.ts
```diff
@@ -0,0 +1,45 @@
...
+    dimension: Promise<Dimension>;
+
+    @CreateDateColumn({ name: 'created_at' })
+    createdAt: Date;
+
+    // This column should be the same across all languages
+    // If one is updated and the others aren't then mark as needing translation
+    @UpdateDateColumn({ name: 'updated_at', nullable: true })
+    updatedAt: Date;
+}
```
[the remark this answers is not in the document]
j-maynard: Also removed this

test/helpers/test-data-source.ts
j-maynard: And removed this... Sorry I must have missed these when I was rebasing

wheelsandcogs (approved):
🚢

PR #18, opened 2024-09-19
Update eslint config and enable shopify ts rules

wheelsandcogs:
This updates the eslint config to use the newer 'flat' config style, turns on the shopify typescript rules and fixes any issues resulting from that.

It resolves the issue we were having with eslint not liking enums.

I've disabled `@typescript-eslint/member-ordering` as it would involve reordering all the static methods in most of the DTOs and ain't nobody got time for that.


eslint.config.mjs
```diff
@@ -0,0 +1,40 @@
...
+    rules: {
+      'line-comment-position': 'off',
+      'no-process-env': 'warn',
+      'no-warning-comments': 'warn',
+      '@typescript-eslint/member-ordering': 'warn',
+      '@typescript-eslint/naming-convention': [
+        'error',
+        {
+          selector: 'default',
+          format: ['camelCase', 'PascalCase', 'UPPER_CASE', 'snake_case'],
```
wheelsandcogs, line 35: default seems to be `['camelCase', 'PascalCase', 'UPPER_CASE']` which gives us a lot of errors due to the DTOs.

src/config/env.enum.ts
```diff
@@ -1,9 +1,7 @@
-/* eslint-disable no-shadow */
-/* eslint-disable no-unused-vars */
```
wheelsandcogs, line 2: Enums work fine now without these.

src/dtos/dataset-dto.ts
```diff
@@ -33,7 +33,7 @@ export class DatasetDTO {
     static async fromDatasetShallow(dataset: Dataset): Promise<DatasetDTO> {
         const dto = new DatasetDTO();
         dto.id = dataset.id;
-        dto.creation_date = dataset.creationDate.toISOString();
+        dto.created_at = dataset.createdAt.toISOString();
```
wheelsandcogs, line 36: all timestamp fields are now `xxx_at`

src/entities/dimension.ts
```diff
@@ -22,7 +19,7 @@ export class Dimension extends BaseEntity {
     @JoinColumn({ name: 'dataset_id', foreignKeyConstraintName: 'FK_dimension_dataset_id' })
     dataset: Promise<Dataset>;
 
-    @Column({ type: 'enum', enum: Object.keys(DimensionType), nullable: false })
+    @Column({ type: 'enum', enum: Object.values(DimensionType), nullable: false })
```
wheelsandcogs, line 22: Everywhere else we were using the enum values rather than enum keys so changed this for consistency.

wheelsandcogs (commented):
This is a fairly dull PR to bring consistency to our naming of things.

j-maynard (approved):
LGTM 🚀

PR #22, opened 2024-09-25
regen migration file to match latest entities

wheelsandcogs:
The current migration file is not consistent with current entity state, so this PR regens it from a blank database and should now exactly match the entities.


j-maynard (approved):
LGTM 🚀

PR #23, opened 2024-09-25
FIX: add missing azure env var

wheelsandcogs:
Missed this in the config PR.


azure-pipelines.yml
```diff
@@ -49,4 +49,33 @@ stages:
...
+          AUTH_PROVIDERS=$(auth_providers)
+          GOOGLE_CLIENT_ID=$(google_client_id)
+          GOOGLE_CLIENT_SECRET=$(google_client_secret)
+          ONELOGIN_URL=$(onelogin_url)
+          ONELOGIN_CLIENT_ID=$(onelogin_client_id)
+          ONELOGIN_CLIENT_SECRET=$(onelogin_client_secret)
+          ONELOGIN_PUBLIC_KEY=$(onelogin_public_key)
+          ONELOGIN_PRIVATE_KEY=$(onelogin_private_key)
+          REDIS_URL=$(redis_url)
+          REDIS_ACCESS_KEY=$(redis_access_key)
```
wheelsandcogs, line 81: This should return as one line without newlines, but will make the git diffs a lot easier to read.

PR #25, opened 2024-09-26
Updated devops pipelines to stop silently failing builds - the new pi…

easternbloc:
…peline will also roll back to the previous green version if it fails


wheelsandcogs:
:shipit:

PR #29, opened 2024-10-03
SW-209: use header instead of path for language

wheelsandcogs:
This removes the current language from URL paths and uses the `Accept-Language` header instead.


package.json
```diff
@@ -5,13 +5,12 @@
   "main": "server.js",
   "scripts": {
     "clean": "rimraf dist/*",
-    "copy-assets": "ts-node tools/copy-assets",
```
wheelsandcogs, line 8: There are no assets to copy on the backend.

src/controllers/auth.ts
```diff
@@ -46,6 +47,8 @@ export const loginGoogle: RequestHandler = (req, res, next) => {
 export const loginOneLogin: RequestHandler = (req, res, next) => {
     logger.debug('attempting to authenticate with one-login...');
 
+    const returnURL = `${config.frontend.url}/${req.language}/auth/callback`;
```
wheelsandcogs, line 50: Currently this will always be the browser language and not the language sent from the frontend (because it's been to the Auth provider and back before it hits this route).

I will fix this in an upcoming PR (using the session, a cookie, or some other method to remember the chosen language).

src/middleware/translation.ts
```diff
@@ -2,19 +2,20 @@ import i18next from 'i18next';
...
     .use(Backend)
     .use(i18nextMiddleware.LanguageDetector)
     .init({
         detection: {
-            order: ['path', 'header'],
+            order: ['header'],
             lookupHeader: 'accept-language',
+            ignoreCase: true,
             caches: false,
-            ignoreRoutes: ['/healthcheck', '/public', '/css', '/assets']
```
wheelsandcogs, line 17: There are no public/css/assets routes on the backend.

j-maynard (approved):
LGTM 🚀

PR #30, opened 2024-10-07
Various language improvements

wheelsandcogs:
A few small changes around language handling:
  * Locales are now defined as an enum
  * Supported languages are loaded from config so that we can change the language support based on environment


easternbloc (approved):
LGTM

PR #31, opened 2024-10-09
Seeder and Reference Data

j-maynard:
This PR adds the reference data in to our code base and a way to load that seed data in to our database.  It also provides entities for us to access the reference data in our code base.


src/reference-data-entities/category.ts
```diff
@@ -0,0 +1,7 @@
+import { Entity, BaseEntity, PrimaryColumn } from 'typeorm';
+
+@Entity({ name: 'categories' })
```
wheelsandcogs, line 3: Why plural here?

j-maynard: Its what was in the original SQL and I went with it.  Figured this was Owains area of expertise and I wouldn't meddle to much.

package.json
```diff
@@ -11,21 +11,23 @@
     "lint:ci": "eslint .",
     "lint:fix": "eslint . --fix",
     "build": "npm-run-all clean tsc",
-    "pretest-podman": "podman compose up -d db-test",
-    "pretest": "docker compose up -d db-test",
+    "pretest": "f() { if [ -x $(command -v podman) ]; then podman compose up -d db-test; else docker compose up -d db-test; fi }; f",
```
wheelsandcogs, line 14: 👌

src/entities/dataset/user.ts
```diff
@@ -1,6 +1,6 @@
 import { BaseEntity, Column, CreateDateColumn, Entity, Index, PrimaryGeneratedColumn, UpdateDateColumn } from 'typeorm';
 
-@Entity({ name: 'users' })
+@Entity({ name: 'user' })
```
wheelsandcogs, line 3: 🫶

wheelsandcogs (approved):
LGTM - thanks for humouring my pedantry :)

PR #39, opened 2024-10-21
Updated the backend to process CSV straight to SQLite3

j-maynard:
This updates the backend to copy the csv data directly to an SQLite3 file which is much easier to query and much faster.  This allows us to get SQLite3 to handle things like pagination, pulling out small parts of the facts table or single columns as needed for the source preview.


src/controllers/csv-processor.ts
```diff
@@ -381,15 +434,17 @@ export const processCSVFromBlobStorage = async (
             dataset_id: dataset.id
         };
     }
-    return processCSVData(buff, page, size, dataset, importObj);
+    return processSQLiteData(buff, page, size, dataset, importObj);
 };
```
wheelsandcogs, line 438: The functions `processSQLiteData`, `getColumnPreview`, `getFileFromDataLake`, etc seem to be unrelated to CSVs now - we should probably look at splitting this stuff out into a SQLite/cube specific module in future.

wheelsandcogs (approved):
Couple of small suggestions.

PR #47, opened 2024-11-15
Owain jones patch 2

OwainJones:
updating to reflect changes/additions to reference data


j-maynard (approved):
LGTM 🚀

OwainJones:
Added in hierarchies for communities first areas. Made updated and additions to LSOAs so we have the 2001, 2011 and 2021 versions. Did the same for MSOAs so we have both 2011 and 2021 vesrions (none of the datasets in SW2 use the 2001 versions so I don;t think we need them). Also added necessary hierarchies for LSOA/MSOA to LA (and to Wales). Made a few tweaks to age data to add three items, 0-17 and 18-64 categories for Broad age groupings, and a 75+ category for the five year age bands. Mainly to match with what is used in the wimd data.

PR #50, opened 2024-11-28
Use type text instead of ltree for hierarchy path

wheelsandcogs:
The `ltree` type is available out of the box in the default Postgres docker image, but not in Azure. We could enable it through the extensions option in the settings, but we can do without it for this particular use-case.


j-maynard (approved):
LGTM 🚀

PR #52, opened 2024-12-02
SW-162: Add route to update revision publish date

wheelsandcogs:
Adds a new route for updating the publish date of a revision.

Includes basic validation for the time being:
 * Revision is not already approved
 * Date is valid and in the future


j-maynard (approved):
LGTM 🚀 

Though adding date-fns is going to cause me a merge conflict or two 😆

PR #54, opened 2024-12-09
SW-321: Dataset Team / Org

wheelsandcogs:
Allows an organisation and team to be selected for the dataset. Once we have user teams built out we can just default to the user's team (or filter the list if they're in more than one team).

Screenshot 2024-12-09 at 16 23 46


j-maynard (approved):
LGTM 🚀

PR #62, opened 2024-12-19
extra check for values

wheelsandcogs:
Check that both sets of translations are populated and throw error if not.


j-maynard (approved):
LGTM 🚀

PR #63, opened 2024-12-19
Added measure handling and major refactor to dataset router

j-maynard:
The intention of this commit is to start the process of trying to reduce some of the code duplication and to make the code a little easier to read and maintain.  The cube handler has be refactored in to smaller functions.  Exported interfaces and functions which are used in a number of places have been moved out in to files of their own and there's now an extractors folder for the extractors used by dimensions and measure.

Also added is the measure handling code to allow the uploading of measures.  The intention is to proper measure handling and note code handling in to the cube handler.


src/controllers/csv-processor.ts
```diff
@@ -155,6 +157,20 @@ export async function extractTableInformation(fileBuffer: Buffer, fileType: File
...
 
+function convertBufferToUTF8(buffer: Buffer): Buffer {
+    const fileEncoding = detectCharacterEncoding(buffer)?.encoding;
+    if (!fileEncoding) {
+        logger.warn('Could not detect file encoding for the file');
+        throw new Error('errors.csv.invalid');
+    }
+    if (fileEncoding !== 'UTF-8') {
+        logger.warn(`File is not UTF-8 encoded... File appears to be ${fileEncoding}... Going to try to recode it`);
+        const decodedString = iconv.decode(buffer, fileEncoding);
```
wheelsandcogs, line 168: Am I right in thinking this is storing the entire CSV data in both `buffer` and `decodedString`?

We might want to look at a streaming solution post-MVP as if they're both several hundred MB we might start seeing memory issues on the deployed containers.

j-maynard: Good point... Yes it is.... We need to dump the original buffer once the conversation is done

src/controllers/revision.ts
```diff
@@ -0,0 +1,478 @@
...
+        fs.unlinkSync(downloadFile);
+        res.end('Server Error');
+    });
+
+    // Optionally listen for the end of the stream
+    downloadStream.on('end', () => {
+        fs.unlinkSync(downloadFile);
+        logger.debug('File stream ended');
+    });
+};
```
wheelsandcogs, line 431: There are very similar handlers above for downloading cubes - do we need both sets?

src/dtos/dataset-dto.ts
```diff
@@ -24,7 +26,9 @@ export class DatasetDTO {
     datasetInfo: DatasetInfoDTO[];
     providers: DatasetProviderDTO[];
     topics: DatasetTopicDTO[];
-    team_id?: string;
+    team?: TeamDTO[];
+    start_date?: Date | null;
+    end_date?: Date | null;
```
wheelsandcogs, line 31: Do these need to be `| null` rather than just optional?

src/dtos/lookup-patch-dto.ts
```diff
@@ -0,0 +1,7 @@
+export interface LookupTablePatchDTO {
+    join_column: string;
+    description_columns: string[];
+    notes_column: string[] | undefined;
+    sort_column: string | undefined;
+    hierarchy: string | undefined;
```
wheelsandcogs, line 6: Prefer `hierarchy?: string` over `hierarchy: string | undefined`

src/dtos/measure-dto.ts
```diff
@@ -1,17 +1,22 @@
...
 
 import { LookupTableDTO } from './lookup-table-dto';
+import { MeasureInfoDTO } from './measure-info-dto';
 
 export class MeasureDTO {
     id: string;
     dataset_id: string;
     fact_table_column: string;
-    join_column: string;
+    join_column: string | null;
```
wheelsandcogs, line 10: Can this just be `join_column?: string` ?

src/dtos/measure-lookup-patch-dto.ts
```diff
@@ -0,0 +1,10 @@
+import { ColumnDescriptor } from '../extractors/column-descriptor';
+
+export interface MeasureLookupPatchDTO {
+    join_column: string;
+    description_columns: string[];
+    notes_columns: string[] | undefined;
+    sort_column: string | undefined;
```
wheelsandcogs, line 7: As above, `sort_column?: string`

src/entities/dataset/dataset.ts
```diff
@@ -39,6 +39,12 @@ export class Dataset extends BaseEntity {
     @Column({ type: 'timestamptz', nullable: true })
     archive: Date;
 
+    @Column({ name: 'start_date', type: 'date', nullable: true })
+    startDate: Date | null;
+
+    @Column({ name: 'end_date', type: 'date', nullable: true })
+    endDate: Date | null;
```
wheelsandcogs, line 46: As above, do we need the `| null` or will `startDate?: Date` suffice?

src/extractors/lookup-table-extractor.ts
```diff
@@ -0,0 +1,8 @@
+import { ColumnDescriptor } from './column-descriptor';
+
+export interface LookupTableExtractor {
+    descriptionColumns: ColumnDescriptor[];
+    sortColumn: string | undefined;
+    hierarchyColumn: string | undefined;
+    notesColumns: ColumnDescriptor[] | undefined;
```
wheelsandcogs, line 7: As above, prefer `notesColumns?: ColumnDescriptor[];`

src/utils/paging-validation.ts
```diff
@@ -0,0 +1,78 @@
...
+
+function validatePageSize(page_size: number): boolean {
+    return !(page_size > MAX_PAGE_SIZE || page_size < MIN_PAGE_SIZE);
+}
+
+function validatePageNumber(page_number: number): boolean {
+    return page_number >= 1;
+}
+
+function validatMaxPageNumber(page_number: number, max_page_number: number): boolean {
```
wheelsandcogs, line 15: typo

wheelsandcogs (approved):
We might want to fix the simple stuff now while it's fresh - there are some larger concerns I have about certain things (see below) but lets get this merged so we can do some user testing.


Possible concerns:
  * We're doing a lot of storing of potentially quite large (100MB+?) datasets in buffers / strings, i.e. in server memory. We probably want to minimise this as much as possible and use streams where we can. It's probably quick at the minute in single user mode, but once we start having multiple users, a gig of container memory is going to disappear pretty quick if there's several datasets being handled at the same time.

  * There seems to be a lot of writing to temp files on GET (i.e. READ operations) - is this something we can maybe shift to being done on Write? e.g. when things are changed, generate all the necessary bits once, and then every Read Op is mostly just streaming data from the DataLake / DB?

src/controllers/cube-handler.ts
```diff
@@ -654,3 +849,224 @@ export const outputCube = async (cubeFile: string, lang: string, mode: DuckdbOut
...
+    if (!latestRevision) {
+        next(new UnknownException('errors.no_revision'));
+        return;
+    }
+    let cubeFile: string;
+    if (latestRevision.onlineCubeFilename) {
+        const dataLakeService = new DataLakeService();
+        const fileBuffer = await dataLakeService.getFileBuffer(latestRevision.onlineCubeFilename, dataset.id);
+        cubeFile = tmp.tmpNameSync({ postfix: '.duckdb' });
+        fs.writeFileSync(cubeFile, fileBuffer);
```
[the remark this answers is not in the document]
j-maynard: Nope you're not missing something... its because I wrote code to handle CSVs before I wrote code to handle a whole cube.  My working assumption is largely there are no cubes in the datalake yet so we need to create the cube and send that... Let me fix this.

src/controllers/cube-handler.ts
```diff
@@ -460,32 +649,111 @@ export const createBaseCube = async (dataset: Dataset, endRevision: Revision): P
...
     }
-    logger.debug(`Writing memory database to disk`);
     const tmpFile = tmp.tmpNameSync({ postfix: '.db' });
-    logger.debug(`Writing memory database to disk at ${tmpFile}`);
-    await quack.exec(`ATTACH '${tmpFile}' as outDB;`);
-    await quack.exec(`COPY FROM DATABASE memory TO outDB;`);
-    await quack.exec('DETACH outDB;');
+    try {
+        logger.debug(`Writing memory database to disk`);
+        logger.debug(`Writing memory database to disk at ${tmpFile}`);
```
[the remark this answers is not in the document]
j-maynard: Removed line 748

src/controllers/cube-handler.ts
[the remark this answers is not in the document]
wheelsandcogs: @j-maynard let's do a fix in another pass, it's not vital right now. See my other comment about potentially moving the file generation to on change rather than on read

src/controllers/lookup-table-handler.ts
```diff
@@ -2,47 +2,127 @@ import fs from 'fs';
...
+                        .map((info) => columnIdentification(info))[0]
+            ),
+            notesColumns: tableMatcher.notes_column?.map(
+                (desc) =>
+                    protoLookupTable.factTableInfo
+                        .filter((info) => info.columnName === desc)
+                        .map((info) => columnIdentification(info))[0]
+            )
+        };
+    } else {
```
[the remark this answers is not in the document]
j-maynard: No the table matcher can be undefined.  The preferred route is nearly always the `else` route.  This is where we try to workout which column is which based on the lookup table.  We should only need a matcher if the user has given us something that doesn't quite match the guidance.  e.g. has extra columns in like the Geography lookup supplied by health.

src/controllers/lookup-table-handler.ts
```diff
@@ -71,60 +154,51 @@ export const validateLookupTable = async (
...
-            if (info.columnName.toLowerCase().startsWith('note')) return false;
+            if (info.columnName.toLowerCase().indexOf('description') > -1) return false;
+            if (info.columnName.toLowerCase().indexOf('sort') > -1) return false;
+            if (info.columnName.toLowerCase().indexOf('hierarchy') > -1) return false;
+            if (info.columnName.toLowerCase().indexOf('note') > -1) return false;
             return true;
         });
         if (possibleJoinColumns.length > 1) {
+            logger.error(
+                `There are to many possible join columns.  Ask user for more information.  Join Columns found: ${JSON.stringify(possibleJoinColumns)}`
```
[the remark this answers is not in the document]
j-maynard: fixed

src/controllers/measure-handler.ts
```diff
@@ -0,0 +1,574 @@
...
+        return;
+    }
+    try {
+        const preview = await getMeasurePreview(dataset, factTable);
+        res.status(200);
+        res.json(preview);
+    } catch (err) {
+        logger.error(`Something went wrong trying to get a preview of the dimension with the following error: ${err}`);
+        res.status(500);
+        res.json({ message: 'Something went wrong trying to generate a preview of the dimension' });
```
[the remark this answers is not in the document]
j-maynard: We can... Changed it.

src/controllers/revision.ts
j-maynard: Yes... One relates to downloading the latest version of the cube... These relate to downloading a cube from a specific revision.  e.g. You're working on a revision which isn't published and want a preview of it or you need a historic revision of a cube.

You'll always need both... Sometimes you just want the latest version off the dataset.. publishers will mostly want one based on their current revision.  The current revision means it doesn't need to be published vs off the dataset where there *needs* to be a revision index to include it in the cube output.

src/dtos/dataset-dto.ts
j-maynard: Yes... Where I've added "null" its because you can't just set these to undefined for TypeORM to take the entries out of the database.  Things should generally exist or be `null` for TypeORM and for JavaScript Null and Undefined are of course different states of being.  So where I've added `| null` its because I need to be able to unset these values.

src/dtos/lookup-patch-dto.ts
j-maynard: Agreed... Let me fix that.

src/dtos/measure-dto.ts
j-maynard: No because in typescript null is different to undefined and join column can be `null`.  This happens after you've done a reset on the measure dimension.

src/controllers/lookup-table-handler.ts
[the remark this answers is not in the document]
wheelsandcogs: What I mean is...
```
if (foo?.bar) {
  return 'x';
} else {
  return 'y';
}
```
is the same as
```
if (foo?.bar) {
  return 'x';
}

return 'y';
```

src/dtos/measure-lookup-patch-dto.ts
j-maynard: Fixed

src/entities/dataset/dataset.ts
j-maynard: You need the null to be able unset the value.  I think this is an issue around TypeScript and of course undefined and null being different states in JavaScript.

src/extractors/lookup-table-extractor.ts
j-maynard: Fixed

src/utils/paging-validation.ts
j-maynard: Fixed

PR #67, opened 2025-01-07
Implement Reference Data

j-maynard:
Continues the refactor of code to reduce duplicates.  Fixes a number of bugs with lookup table and measure lookup table matching and makes sure that all CSVs uploaded destined for DuckDB and now converted to UTF-8 before being passed through.

The key feature is the implementation of basic reference data with a little bit of reference data validation which can be expanded on post MVP.


src/services/csv-processor.ts
```diff
@@ -22,6 +21,9 @@ import { FileType } from '../enums/file-type';
 import { FactTableInfo } from '../entities/dataset/fact-table-info';
 import { FactTableAction } from '../enums/fact-table-action';
 
+import { DataLakeService } from './datalake';
+import { convertBufferToUTF8 } from '../utils/file-utils';
```
wheelsandcogs: 👍

src/services/cube-handler.ts
```diff
@@ -726,8 +867,20 @@ export const createBaseCube = async (dataset: Dataset, endRevision: Revision): P
...
+    }
+
     await setupDimensions(quack, dataset, selectStatementsMap, joinStatements, orderByStatements);
 
+    if (referenceDataPresent(dataset)) {
+        await cleanUpReferenceDataTables(quack);
+        joinStatements.push(`JOIN reference_data_info ON reference_data.item_id=reference_data_info.item_id`);
+        joinStatements.push(`    AND reference_data.category_key=reference_data_info.category_key`);
+        joinStatements.push(`    AND reference_data.version_no=reference_data_info.version_no`);
+        joinStatements.push(`    AND reference_data_info.lang='#LANG#'`);
```
wheelsandcogs, line 881: This whitespace is a bit strange... is this to make it more readable in debug logs?

wheelsandcogs (approved):

src/services/cube-handler.ts
j-maynard: Yes... When you get the debug statement for the default view its to allow picking out the reference data query part of the query.

PR #68, opened 2025-01-10
SW-324: Publisher homepage dataset list

wheelsandcogs:
This adds the necessary properties to the dataset list query in order to display the dataset list on the publisher homepage.


j-maynard (approved):
LGTM 🚀

PR #69, opened 2025-01-10
SW-249: List published datasets for consumer side

wheelsandcogs:
This adds a new unauthed route to fetch a paginated list of published datasets.


j-maynard (approved):
LGTM 🚀

PR #70, opened 2025-01-13
FIX: Actually calculate the offset

wheelsandcogs:
Forgot to actually calculate the offset from the page number and items per page.


j-maynard (approved):
LGTM 🚀

PR #72, opened 2025-01-14
Fix missing values when measure is incomplete

j-maynard:
This fixes a bug where missing values happen in the cube when there is an issue with the measure lookup.  The fix adds an else clause to the case statement used to display the data values.


wheelsandcogs (approved):
🚢

PR #73, opened 2025-01-14
Casade dataset deletes to topics to prevent error

wheelsandcogs:
If you try and manually delete datasets from the database that have topics, it will error on FK.


easternbloc (approved):
LGTM

PR #78, opened 2025-01-17
Allow email to be used with multiple providers

wheelsandcogs:
This removes the error when a different auth provider is used to log in with the same email address.

We now just update our local user record with the latest provider details.


easternbloc (approved):
LGTM. Long term it's not really going to be an issue as it'll only be Entra...

PR #81, opened 2025-01-21
SW-264: capture an audit log of changes to the database and who made them

wheelsandcogs:
For auditing purposes we need an event log to track who is making changes to the datasets we have and when they made them.

Screenshot 2025-01-21 at 17 49 45


j-maynard (approved):
LGTM 🚀

PR #83, opened 2025-01-24
SW-337: submit dataset for publication

wheelsandcogs:
For MVP, this sets the `revision.approvedAt` date so that we can test the "scheduled" and "published" statuses.

Post-MVP, we will have an approval stage and `approvedAt` will not be set until that journey is completed.


j-maynard (approved):
LGTM 🚀

PR #84, opened 2025-01-27
SW-338: withdraw revision from publication

wheelsandcogs:
This adds a new route to allow a scheduled publication to be withdrawn.

For MVP, this just involves nullifying the `approvedAt` and `approvedBy` fields on the scheduled revision.


j-maynard (approved):
LGTM 🚀

PR #86, opened 2025-01-30
Check all required config is present

wheelsandcogs:


src/config/app-config.interface.ts
```diff
@@ -77,3 +77,7 @@ export interface AppConfig {
         };
     };
 }
+
+// list any optional properties here so we can ignore missing values when we check the config on boot
+// it would be nice to get them directly from the interface, but interfaces are compile-time only
+export const optionalProperties = ['redisUrl', 'redisPassword', 'google', 'entraid', 'blob', 'datalake'];
```
wheelsandcogs, line 83: These are optional because they are not used in every env (e.g. CI) - it should be obvious once the app tries to connect if the connection config is missing.

easternbloc (approved):

PR #88, opened 2025-01-31
SW-429: fix list status for scheduled datasets

wheelsandcogs:
Scheduled datasets were showing as "Live" before the live date was in the past.


easternbloc (approved):
LGTM

PR #91, opened 2025-02-03
Fix bug with specific dates

wheelsandcogs:


src/services/dimension-processor.ts
```diff
@@ -343,16 +344,13 @@ export const validateDateTypeDimension = async (
         monthFormat: dimensionPatchRequest.month_format,
         dateFormat: dimensionPatchRequest.date_format
     };
-    logger.debug(`Extractor created with the following JSON: ${JSON.stringify(extractor)}`);
+    logger.debug(`Extractor created with: ${JSON.stringify(extractor)}`);
     const previewQuery = `SELECT DISTINCT "${dimension.factTableColumn}" FROM ${tableName}`;
     const preview = await quack.all(previewQuery);
     try {
-        logger.debug(`Extractor created with ${JSON.stringify(extractor)}`);
```
wheelsandcogs, line 350: Removed duplicate debug message

src/services/time-matching.ts
```diff
@@ -239,58 +239,66 @@ function periodTableCreator(dateFormat: DateExtractor, dataColumn: TableData) {
...
         }
 
         referenceTable.push({
-            dateCode: value.toString(),
+            dateCode: value,
             description: format(parsedDate, 'dd/MM/yyyy'),
             start: parsedDate,
             end: sub(add(parsedDate, { days: 1 }), { seconds: 1 }),
-            type: 'specific-day'
+            type: 'specific_day'
```
wheelsandcogs, line 289: This key is only used in the translations as far as I can tell... so made it snake_case like the `${dateFormat.type}_year` ones above.

easternbloc (approved):

PR #97, opened 2025-02-05
Don't fall over if some auth config missing

wheelsandcogs:
Backend will now still boot up if an auth provider is enabled but config is missing.

Does not affect the frontend buttons, they're configured in the frontend.


easternbloc (approved):
LGTM

easternbloc (approved):

PR #105, opened 2025-02-07
Tidying up schema

wheelsandcogs:
Some tidying up -
 * there were some things in the entities that were not represented in the db after running the initial migration. Not sure why but after running the `resolve-outstanding` migration we now get "No changes in database schema were found" when running `npm run migration:generate` which is what we would expect
 * Some of the join tables (dataset_provider, dataset_topic) were missing on delete cascade
 * Some of the FKs around ref data were not named, makes it harder to identify what they're for
 * Both sides of the lookup_table <-> measure and lookup_table <-> dimension joins had an FK to the other side, we only need a single side to own the relationship, otherwise they can get out of sync and or cause other side-effects

There will be a part 2 coming with indexes but didn't want this to grow too large.

### DIFF
--- a/src/controllers/measure-controller.ts
+++ b/src/controllers/measure-controller.ts
@@ -62,8 +62,10 @@ export const attachLookupTableToMeasure = async (req: Request, res: Response, ne
         next(new NotFoundException('errors.fact_table_invalid'));
         return;
     }
+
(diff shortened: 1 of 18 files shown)


j-maynard (approved):
LGTM 🚀

PR #106, opened 2025-02-07
Limit DuckDBs memory usage and thread usage

j-maynard:


src/controllers/cube-controller.ts
```diff
@@ -26,6 +26,8 @@ export const getCubePreview = async (
 ): Promise<ViewDTO | ViewErrDTO> => {
     logger.debug(`Opening cube file ${cubeFile}`);
     const quack = await Database.create(cubeFile);
+    await quack.exec('SET threads = 1;');
+    await quack.exec("SET memory_limit = '15MB';");
```
wheelsandcogs, line 30: Should we be using the config here?

src/controllers/cube-controller.ts
```diff
@@ -83,6 +85,8 @@ export const getCubePreview = async (
 
 export const outputCube = async (cubeFile: string, lang: string, mode: DuckdbOutputType) => {
     const quack = await Database.create(cubeFile);
+    await quack.exec('SET threads = 1;');
+    await quack.exec("SET memory_limit = '15MB';");
```
wheelsandcogs, line 89: Should we be using the config here?

src/controllers/revision.ts
```diff
@@ -272,6 +272,8 @@ async function attachUpdateDataTableToRevision(
 
     revision.dataTable = fileImport;
     const quack = await Database.create(':memory:');
+    await quack.exec('SET threads = 1;');
+    await quack.exec("SET memory_limit = '15MB';");
```
wheelsandcogs, line 276: Should we be using the config here?

wheelsandcogs (approved):

src/controllers/cube-controller.ts
j-maynard: No... we shouldn't... Fixed

src/controllers/cube-controller.ts
j-maynard: No... We shouldn't... Fixed.

src/controllers/revision.ts
j-maynard: Nope... Fixed it.

PR #110, opened 2025-02-09
Improve dataset provider & topic performance

wheelsandcogs:
Adds new routes for fetching the dataset providers /topics so that we don't have to include those relations in the "limited" dataset used by unrelated requests.

https://github.com/Marvell-Consulting/statswales-frontend/pull/121


j-maynard (approved):
LGTM 🚀

PR #112, opened 2025-02-10
fix: include measures in tasklist

wheelsandcogs:


src/controllers/revision.ts
```diff
@@ -431,7 +428,7 @@ export const updateRevisionPublicationDate = async (req: Request, res: Response,
         }
 
         await RevisionRepository.updatePublishDate(revision, publishAt);
-        const updatedDataset = await DatasetRepository.getById(dataset.id);
+        const updatedDataset = await DatasetRepository.getById(dataset.id, {});
```
wheelsandcogs, line 431: A lot of the time the dataset returned to the frontend is only used to verify the action completed and then redirect the user back to the tasklist or overview.

We don't need to join any relations in these cases.

src/route/dataset.ts
```diff
@@ -190,6 +190,7 @@ const relForTasklistState: FindOptionsRelations<Dataset> = {
     metadata: true,
     revisions: { dataTable: true },
     dimensions: { metadata: true },
+    measure: { measureInfo: true },
```
wheelsandcogs, line 193: This fixes the missing measures on tasklist

j-maynard (approved):
lgtm 🚀

PR #113, opened 2025-02-10
SW-446: ignore empty rows in the imported translations

wheelsandcogs:
Also quotes all strings to see if it helps some Excel users who were seeing the cells not populated correctly.


j-maynard (approved):
lgtm 🚀

PR #124, opened 2025-02-25
Limit relations loaded for revision routes

j-maynard:
Specified which relations to load for the revision routes.  Removed the reference data entities and seeder as these are not used in our code currently and the direction of travel means reference data will come via API or be loaded from CSVs in the code base longer term.


src/migrations/1738599982159-initial-schema.ts
```diff
@@ -580,27 +580,6 @@ export class InitialSchema1738599982159 implements MigrationInterface {
...
-        `);
-        await queryRunner.query(`
-            ALTER TABLE "reference_data_info" DROP CONSTRAINT "FK_f671fde9c769286ba971485ba09"
-        `);
-        await queryRunner.query(`
-            ALTER TABLE "reference_data" DROP CONSTRAINT "FK_dd4ff535904e339641b0b0d52c2"
-        `);
-        await queryRunner.query(`
-            ALTER TABLE "category_key" DROP CONSTRAINT "FK_087b36846d67092609821a62756"
-        `);
```
wheelsandcogs, line 603: Please can we avoid editing existing schema files as they've already been run and we'll end up with out-of-sync DBs... Can we add all the required changes to the new migration file? 

Failing that, let's just replace the initial schema completely with a fresh single migration.

j-maynard: Restored the initial migration file

PR #132, opened 2025-02-27
Clean up source implementation to remove everything and recreate

j-maynard:
This addresses ticket SW-382 by changing the behaviour to remove all fact table columns, dimensions and measures and effectivly start the create process from scratch if the user changes either the sources or based fact table.

### DIFF
--- a/src/services/dimension-processor.ts
+++ b/src/services/dimension-processor.ts
@@ -27,6 +27,9 @@ import { getFileImportAndSaveToDisk, loadFileIntoDatabase } from '../utils/file-
 import { LookupTableExtractor } from '../extractors/lookup-table-extractor';
 import { LookupTable } from '../entities/dataset/lookup-table';
 import { FactTableColumn } from '../entities/dataset/fact-table-column';
+import { MeasureRow } from '../entities/dataset/measure-row';
+import { MeasureMetadata } from '../entities/dataset/measure-metadata';
+import { measureRouter } from '../route/measure';
 
 import { DateReferenceDataItem, dateDimensionReferenceTableCreator } from './time-matching';
 import { createFactTableQuery } from './cube-handler';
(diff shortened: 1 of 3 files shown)


src/services/dimension-processor.ts
```diff
@@ -310,14 +229,58 @@ async function recreateBaseFactTable(dataset: Dataset, dataTable: DataTable): Pr
...
+            } catch (error) {
+                logger.warn(
+                    error,
+                    `Something went wrong trying to remove previously uploaded lookup table with error: ${error}`
+                );
+            }
+            logger.warn(`Removing lookup table for dimension ${dimension.id}`);
+            await LookupTable.getRepository().delete({ dimension });
+        }
+        await DimensionMetadata.getRepository().delete({ dimension });
```
wheelsandcogs, line 248: We've got an `onDelete: 'CASCADE'` on the metadata side of this join - shouldn't they get automatically be deleted when you remove the dimension record?

Maybe it only happens with `.remove()` and not `.delete()`? 

We don't have an `onDelete` on the LookupTable - is it worth adding?

src/services/dimension-processor.ts
```diff
@@ -310,14 +229,58 @@ async function recreateBaseFactTable(dataset: Dataset, dataTable: DataTable): Pr
...
+                logger.warn(
+                    error,
+                    `Something went wrong trying to remove previously uploaded lookup table with error: ${error}`
+                );
+            }
+            logger.warn(`Removing lookup table for measure ${dataset.measure.id}`);
+            await LookupTable.getRepository().delete({ measure: dataset.measure });
+        }
+        await MeasureRow.getRepository().delete({ measure: dataset.measure });
+        await MeasureMetadata.getRepository().delete({ measure: dataset.measure });
```
wheelsandcogs, line 270: Similar to above - both MeasureRow and MeasureMetadata have an `onDelete:'CASCADE'` on the measure relation - shouldn't this get triggered automatically by deleting the measure?

src/services/dimension-processor.ts
j-maynard: Yes we probably should although I wonder how it will work as both measure and dimension both use the same lookup table entity.

wheelsandcogs: you can have the same lookup table row referenced by both a dimension and a measure?

PR #133, opened 2025-02-28
SW-308: Move metadata to revision, fix various metadata & translation bugs

wheelsandcogs:
This PR moves metadata, topics and providers from the dataset entity to the revision.

It requires the frontend PR: https://github.com/Marvell-Consulting/statswales-frontend/pull/143

It also fixes the following issues:
 * SW-460: Metadata properties that don't need translation should show as complete for both languages
 * SW-423: Show blank cell for Welsh if dimension name not provided
 * SW-420: Related report link text should be in translation export


j-maynard (approved):
LGTM 🚀

PR #147, opened 2025-03-11
Add support for numeric dimensions

j-maynard:


src/services/dimension-processor.ts
```diff
@@ -202,18 +213,14 @@ async function createUpdateMeasure(dataset: Dataset, columnAssignment: SourceAss
...
-    metadata.language = lang;
-    metadata.name = columnInfo.columnName;
-    await metadata.save();
-  });
+
+  await Measure.create({
+    dataset,
+    factTableColumn: columnInfo.columnName,
+    metadata: SUPPORTED_LOCALES.map((language: string) =>
+      DimensionMetadata.create({ language, name: columnInfo.columnName })
```
wheelsandcogs, line 221: Not MeasureMetadata?

src/services/dimension-processor.ts
```diff
@@ -228,18 +235,17 @@ async function createUpdateNoteCodes(dataset: Dataset, columnAssignment: SourceA
...
     factTableCol.columnDatatype = col.columnDatatype;
     factTableCol.columnIndex = col.columnIndex;
     factTableCol.id = dataset.id;
     factTableCol.dataset = dataset;
-    const savedFactTableCol = await factTableCol.save();
-    factTable.push(savedFactTableCol);
-  }
+    return factTableCol;
+  });
+  await FactTableColumn.save(factTable);
```
wheelsandcogs, line 248: Not a blocker but you could use the create method 

```
const factTable = dataTable.dataTableDescriptions.map((col) => {
  return FactTableColumn.create({
    columnType: FactTableColumnType.Unknown,
    ...etc
  });
});
```

wheelsandcogs (approved):

src/services/dimension-processor.ts
j-maynard: My bad... fixed it.

PR #150, opened 2025-03-12
re-add prettier eslint

wheelsandcogs:
We lost the prettier bit of the eslint checks when removing shopify, so we no-longer get the red squiggly underlines in the editor for prettier issues.

This adds them back in.


j-maynard (approved):
LGTM 🚀

PR #153, opened 2025-03-18
deploy backend on merge to main

easternbloc:


.github/workflows/deploy-backend.yml
```diff
@@ -0,0 +1,46 @@
+name: Deploy Frontend
```
wheelsandcogs, line 1: Frontend?

wheelsandcogs:
Can we call the workflow `deploy-dev` ?

.github/workflows/deploy-backend.yml
easternbloc: Good spot!

wheelsandcogs (approved):

PR #154, opened 2025-03-18
Allow storage to be configurable per env (datalake or blob storage)

wheelsandcogs:
This adds a new interface `StorageService` and updates the datalake and blob storage classes to implement this interface.

For convenience, the file service is added to the request object `req.fileService` via the services middleware.

If you need it outside of a request context, there is a new helper function: `getFileService()`.

Configuration for the storage service is via `config.storage.store`.

If you need the underlying DataLake or BlobStorage service, you can get it via `fileService.getServiceClient()`.


src/config/envs/ci.ts
```diff
@@ -48,15 +49,14 @@ export function getCIConfig(): AppConfig {
       }
     },
     storage: {
+      store: FileStore.Blob,
```
wheelsandcogs, line 52: CI defaults to blob storage.

src/config/envs/default.ts
```diff
@@ -66,12 +67,15 @@ export const getDefaultConfig = (): AppConfig => {
       }
     },
     storage: {
+      store: FileStore.DataLake,
```
wheelsandcogs, line 70: Default for all envs (unless overridden) is DataLake

src/config/envs/local.ts
```diff
@@ -50,6 +51,9 @@ export function getLocalConfig(): AppConfig {
     duckdb: {
       threads: process.env.DUCKDB_THREADS ? parseInt(process.env.DUCKDB_THREADS, 10) : 1,
       memory: process.env.DUCKDB_MEMORY || '125MB'
+    },
+    storage: {
+      store: (process.env.FILE_STORE as FileStore) || FileStore.DataLake
```
wheelsandcogs, line 56: Local is configurable via env var `FILE_STORE=`.

src/config/envs/production.ts
```diff
@@ -6,6 +7,12 @@ import { AppEnv } from '../env.enum';
...
 export function getProductionConfig(): AppConfig {
   return defineConfig({
-    env: AppEnv.Prod
+    env: AppEnv.Prod,
+    auth: {
+      providers: [AuthProvider.EntraId],
+      jwt: {
+        cookieDomain: process.env.BACKEND_URL!.replace('api.', '')
+      }
+    }
```
wheelsandcogs, line 16: This is unrelated but will be needed eventually, so gonna leave it in.

src/config/envs/staging.ts
```diff
@@ -9,9 +9,9 @@ export function getStagingConfig(): AppConfig {
   return defineConfig({
     env: AppEnv.Staging,
     auth: {
-      providers: [AuthProvider.Google, AuthProvider.EntraId, AuthProvider.Local],
+      providers: [AuthProvider.EntraId, AuthProvider.Local],
```
wheelsandcogs, line 12: Remove google from staging env.

src/services/datalake-storage.ts
```diff
@@ -0,0 +1,109 @@
...
+    return this.getDirectoryClient(directory).getFileClient(filename);
+  }
+
+  public async createDirectoryIfNotExists(directory: string): Promise<DirectoryCreateIfNotExistsResponse> {
+    return this.getDirectoryClient(directory).createIfNotExists();
+  }
+
+  public async saveBuffer(filename: string, directory: string, content: Buffer): Promise<FileUploadResponse> {
+    logger.debug(`Uploading file '${filename}' to datalake as buffer`);
+    await this.createDirectoryIfNotExists(directory);
```
wheelsandcogs, line 62: Rather than having to manually create the directory, just create it on the fly if it doesn't exist when saving.

This might mean a few more requests to datalake service but it means we don't have to think about it outside of this class.

j-maynard (approved):
LGTM 🚀

PR #164, opened 2025-03-20
Fix update previews

j-maynard:
Fixes:
- SW-548 Updating - “Page not found” on measures task
- SW-573 Dimension look ups 'page not found' in update user journey

Issue stems from updates not always having a data table to generate a fact table.  Switch fact table generation to using the one in the cube builder.  Follow up tech debt is to reduce repeated code but that means looking hollistically at dimension and previews.

### DIFF
--- a/src/services/dimension-processor.ts
+++ b/src/services/dimension-processor.ts
@@ -31,12 +31,15 @@ import { MeasureRow } from '../entities/dataset/measure-row';
 import { MeasureMetadata } from '../entities/dataset/measure-metadata';
 
 import { dateDimensionReferenceTableCreator, DateReferenceDataItem } from './time-matching';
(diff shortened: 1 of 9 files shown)


src/services/dimension-processor.ts
```diff
@@ -364,22 +366,26 @@ export const validateNumericDimension = async (
...
 
   const tableName = 'fact_table';
+  let endRevision: Revision;
+  if (dataset.draftRevision) {
+    endRevision = dataset.draftRevision;
+  } else if (dataset.publishedRevision) {
+    endRevision = dataset.publishedRevision;
+  } else {
+    throw new Error('No revision present on the dataset');
+  }
```
wheelsandcogs, line 376: See above about `endRevision`

PR #168, opened 2025-03-27
SW-582: Add / edit user groups

wheelsandcogs:
Adds new schema and endpoints for managing user groups.

Assigning users and datasets to groups will come later.


j-maynard (approved):
LGTM 🚀

PR #171, opened 2025-04-01
Enabling validating fact tables for duplicate and null facts

j-maynard:


src/services/fact-table-validator.ts
```diff
@@ -0,0 +1,176 @@
...
+    await loadFileDataTableIntoTable(quack, dataTable, factTableDef, dataTableFile, FACT_TABLE_NAME);
+  } catch (err) {
+    const error = err as FactTableValidationException;
+    logger.error(error, 'Failed to load data table into fact table');
+    try {
+      await loadFileIntoCube(quack, dataTable, dataTableFile, 'data_table');
+    } catch (extractionError) {
+      logger.error(extractionError, 'Failed to extract data from data table.');
+      throw error;
+    }
```
wheelsandcogs, line 141: Please can you walk me through the nested try catch here? 

We're loading the data table into DuckDB, and if it fails we're loading the file into the cube?

Are we using exceptions here to manage normal flow (e.g. kind of an "if cube not exists" trigger) or is this actually an exceptional case that shouldn't really happen?

src/utils/table-data-to-view-table.ts
```diff
@@ -0,0 +1,24 @@
...
+    let sourceType = FactTableColumnType.Unknown;
+    if (tableHeaders[i] === 'line_number') {
+      sourceType = FactTableColumnType.LineNumber;
+    }
+    headers.push({
+      index: i,
+      name: tableHeaders[i],
+      source_type: sourceType
+    });
+  }
```
wheelsandcogs, line 19: Optional alternative - .map()

```suggestion
  const headers = tableHeaders.map((header, idx) => ({
    index: idx,
    name: header,
    sourceType: header  === 'line_number' ? FactTableColumnType.LineNumber : FactTableColumnType.Unknown
  }));
```

src/services/fact-table-validator.ts
j-maynard: So the nested try catch here allows us to try to query the still open cube to find those values which caused problems when trying to load data into the fact table.  They're still database queries which can fail but ultimately all we're trying to do is add more information to the exception to pass on to the frontend.

If these fail it's not the end of the world.  It won't break anything more than is already broken.  However if we can identify the rows which have issues we give the user a better chance at trying to workout why their fact table failed to load.

The nested try catch allows us to try a thing, if it blows up again to log that it blew up again and just return the original error back to the calling function without throwing a new unless exception which isn't caught which could mask the problem the system actually encountered.

src/enums/fact-table-validation-exception-type.ts
```diff
@@ -0,0 +1,10 @@
+export enum FactTableValidationExceptionType {
+  UnknownPresent = 'unknown_present',
```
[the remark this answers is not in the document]
wheelsandcogs: Can we call it `UnknownSource` maybe?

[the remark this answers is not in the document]
j-maynard: Changed this.

src/services/fact-table-validator.ts
j-maynard: Updated the code and added some comments... Hope this is what you want.

wheelsandcogs (approved):

PR #172, opened 2025-04-02
Add downloading all cube data from the data lake as a zip

j-maynard:
This improves raw file downloads for measure, dimension and data tables. The first two were completely missing and are now needed now we don't always have direcrt access to the datalake in the client estate.

Also added the option to pull down everything from the datalake for a given dataset.  This is helpful when trying to understand why a given dataset journey has failed.

Added a route to list all the file imports which made up a given dataset.  Again this provides us with extra diagnositic information for the developer page.

Fianlly fixed a small issue in the cube builder where there was a slightly broken join statement on date dimensions as the date dimension table switch from being monolinguagl to multilingual.

### DIFF
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -35,6 +35,10 @@ import { TopicSelectionDTO } from '../dtos/topic-selection-dto';
 import { getCubePreview } from './cube-controller';
 import { factTableValidatorFromSource } from '../services/fact-table-validator';
 import { FactTableValidationException } from '../exceptions/fact-table-validation-exception';
+import JSZip from 'jszip';
+import { DataLakeFileEntry } from '../interfaces/datalake-file-entry';
+import { StorageService } from '../interfaces/storage-service';
+import { FileImportDto } from '../dtos/file-import';
 
 export const listAllDatasets = async (req: Request, res: Response, next: NextFunction) => {
   try {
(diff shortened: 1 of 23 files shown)


src/controllers/dataset.ts
```diff
@@ -336,3 +340,80 @@ export const getFactTableDefinition = async (req: Request, res: Response) => {
...
+    } else {
+      filename = fileEntry as string;
+    }
+    const originalFilename = datasetFiles.get(filename)?.filename || filename;
+    zip.file(originalFilename, await fileService.loadBuffer(filename, directory));
+  }
+}
+
+function collectFiles(dataset: Dataset): Map<string, FileImportDto> {
+  const files: Map<string, FileImportDto> = new Map<string, FileImportDto>();
```
wheelsandcogs, line 370: More personal preference than anything, but generally you don't need to type hint stuff when it's implicitly typed by the assignment, feels a bit redundant in cases like this.

src/controllers/dataset.ts
```diff
@@ -336,3 +340,80 @@ export const getFactTableDefinition = async (req: Request, res: Response) => {
...
+  });
+  dataset.revisions.forEach((revision) => {
+    if (revision.dataTable) {
+      files.set(revision.dataTable.filename, FileImportDto.fromFileImport(revision.dataTable));
+    }
+  });
+  return files;
+}
+
+export const getEverythingFromDatalake = async (req: Request, res: Response) => {
```
wheelsandcogs, line 393: I'd prefer to avoid using the word "datalake" or "blobstorage" outside of their respective storage classes...

We've now got `req.fileService` which might be an instance of blob storage or datalake, and this should ideally (if we want to be able to test this endpoint in CI) work with either.

e.g. `getAllFilesForDataset` / `fullDatasetZip` or similar.

src/controllers/dataset.ts
```diff
@@ -336,3 +340,80 @@ export const getFactTableDefinition = async (req: Request, res: Response) => {
...
+  });
+  zip
+    .generateNodeStream({ type: 'nodebuffer', streamFiles: true })
+    .pipe(res)
+    .on('finish', () => {
+      res.end();
+    });
+};
+
+export const listAllFilesInDataset = async (req: Request, res: Response) => {
```
wheelsandcogs, line 421: 👍

src/entities/dataset/lookup-table.ts
```diff
@@ -26,6 +26,9 @@ export class LookupTable extends BaseEntity implements FileImportInterface {
   @Column({ type: 'varchar', length: 255 })
   filename: string;
 
+  @Column({ name: 'original_filename', type: 'varchar', length: 255 })
```
wheelsandcogs, line 29: just use `type: text` (it's basically identical in Postgres and you don't need to worry about length)

src/services/datalake-storage.ts
```diff
@@ -93,15 +94,15 @@ export default class DataLakeStorage implements StorageService {
...
     const files = await this.fileSystemClient.listPaths({ path: directory });
-    const fileList = [];
+    const fileList: DataLakeFileEntry[] = [];
 
     for await (const file of files) {
       if (file.name === undefined) {
         continue;
       }
-      fileList.push({ name: basename(file.name), path: file.name, isDirectory: file.isDirectory });
+      fileList.push({ name: basename(file.name), path: file.name, isDirectory: file.isDirectory ?? false });
```
wheelsandcogs, line 105: Does `file.isDirectory` sometimes return undefined?

wheelsandcogs (changes requested):

src/entities/dataset/lookup-table.ts
j-maynard: Yeah but here I want to worry about length... Windows still has a 255 character limit on filenames and while it's unlikely someone might violate that it's still a possibility.  Also I might upload a file from my Mac which is then attempted to be downloaded to a Windows machine.

src/services/datalake-storage.ts
j-maynard: According to the type hinting in the class... I thought it was weird to but here we are.  Not our code sadly so we just have to roll with it.

src/controllers/dataset.ts
```diff
@@ -336,3 +340,80 @@ export const getFactTableDefinition = async (req: Request, res: Response) => {
...
+      files.set(dimension.lookupTable.filename, FileImportDto.fromFileImport(dimension.lookupTable));
+    }
+  });
+  dataset.revisions.forEach((revision) => {
+    if (revision.dataTable) {
+      files.set(revision.dataTable.filename, FileImportDto.fromFileImport(revision.dataTable));
+    }
+  });
+  return files;
+}
```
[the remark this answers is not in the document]
j-maynard: Moved to utils

src/controllers/dataset.ts
j-maynard: Removed the type hint from this line... My past as a Java Dev came through 😆

src/controllers/dataset.ts
j-maynard: Renamed to `getAllFilesForDataset`

wheelsandcogs (approved):

PR #174, opened 2025-04-03
SW-638 - mark stale translations exports as incomplete

joefitter:
* Updated GET translation/:dataset/export route to store exported translations in `data.translations`
* Abstracted collectTranslations helper
* Updated tasklist-state-dto existingTranslations to check latest exported translation EN values against current dataset. Mark export as incomplete if any are stale

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ yarn.lock
 
 # Ignore certs
 certs/*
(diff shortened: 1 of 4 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #175, opened 2025-04-04
Update nvmrc

j-maynard:
Updates the .nvmrc file to match the one on the frontend


wheelsandcogs (approved):

wheelsandcogs:
I don't think we really need to be specifying specific minor versions, but 🤷

PR #177, opened 2025-04-04
Chore - remove baseUrl from tsconfig

joefitter:
This is to be used in conjunction with `paths` to set a module resolution context, but is only recommended for AMD module loaders in the browser. Leaving it means my IDE defaults to autocompleting paths incorrectly.


j-maynard (approved):
LGTM 🤘🏻

PR #180, opened 2025-04-08
Fix developer view bugs

j-maynard:
Fixes some issues with the developer view.


src/utils/dataset-controller-utils.ts
```diff
@@ -31,7 +31,7 @@ export const addDirectoryToZip = async (
 
 export const collectFiles = (dataset: Dataset): Map<string, FileImportDto> => {
   const files = new Map<string, FileImportDto>();
-  if (dataset.measure.lookupTable) {
+  if (dataset.measure && dataset.measure.lookupTable) {
```
wheelsandcogs, line 34:
```suggestion
  if (dataset.measure?.lookupTable) {
```
... or use optional chaining.

wheelsandcogs (approved):

PR #182, opened 2025-04-08
missed removing a not null

wheelsandcogs:
I've already run this on dev db directly, but it can run again without breaking anything.

This is to make sure our other envs get it as well (preview, your localhost, etc)


j-maynard (approved):
LGTM 🤘🏻

PR #183, opened 2025-04-08
SW-583: View users in a group

wheelsandcogs:
Updates the `GET /admin/group/:groupId` endpoint to include a list of users assigned to the group.


j-maynard (approved):
LGTM 🤘🏻

PR #184, opened 2025-04-09
ignore whitespace around delimiters

wheelsandcogs:
Currently we only handle translation CSVs where the delimiters have no white-space, e.g.:
```
"foo","bar"
```

but not
```
"foo", "bar"
```

This fixes the import for those csvs


j-maynard (approved):
LGTM 🤘🏻

PR #187, opened 2025-04-10
Switch docker-compose and predev to use Valkey

j-maynard:
As we're moving towards using Valkey in our environments its probably a good idea that we're testing and developing against Valkey locally. This PR just updates the docker-compose and package.json to use Valkey instead of Redis.


joefitter (approved):

wheelsandcogs:
We'll need to update the frontend compose/npm scripts as well 👍

PR #188, opened 2025-04-10
Fixes unquoted columns inserting into fact table

j-maynard:


wheelsandcogs (approved):

wheelsandcogs:
This seems to keep happening - can we create a helper function to do this and use it everywhere we're building cube queries?

PR #192, opened 2025-04-11
Fix protocube saving

j-maynard:
It appears then when you do `ducked.close()` it returns before ducked finishes cleaning up its temp which can result in an empty duckdb file being written to the data lake.  This is especially important as we now partially build the cube once the column sources stage has been completed and use this protocube for previews and and dimension/measure validation.

### DIFF
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -330,7 +330,7 @@ export const updateSources = async (req: Request, res: Response, next: NextFunct
     });
     return;
   }
-  let duckdbFile;
+  let duckdbFile: string;
   try {
     duckdbFile = await factTableValidatorFromSource(dataset, validatedSourceAssignment);
   } catch (err) {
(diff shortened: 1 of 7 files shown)


src/controllers/dataset.ts
```diff
@@ -354,10 +354,12 @@ export const updateSources = async (req: Request, res: Response, next: NextFunct
     return;
   }
   try {
+    logger.debug(`Saving duckdb file to blob storage: ${duckdbFile}`);
     const buffer = fs.readFileSync(duckdbFile);
+    logger.debug(`Buffer size: ${buffer.byteLength} bytes`);
     await req.fileService.saveBuffer(`${revision.id}-protocube.duckdb`, dataset.id, buffer);
     revision.onlineCubeFilename = `${revision.id}-protocube.duckdb`;
-    fs.unlinkSync(duckdbFile);
+    //fs.unlinkSync(duckdbFile);
```
wheelsandcogs, line 362: intentional?

src/services/cube-handler.ts
```diff
@@ -1303,32 +1304,12 @@ export const createBaseCube = async (datasetId: string, endRevisionId: string):
...
-  // Pass the file handle to the calling method
-  // If used for preview you just want the file
-  // If it's the end of the publishing step you'll
-  // want to upload the file to the data lake.
   const end = performance.now();
   const functionTime = Math.round(end - functionStart);
   const buildTime = Math.round(end - buildStart);
   logger.warn(`Cube function took ${functionTime}ms to complete and it took ${buildTime}ms to build the cube.`);
+  await quack.close();
+  await new Promise((f) => setTimeout(f, DUCKDB_WRITE_TIMEOUT));
```
wheelsandcogs, line 1312: can we make this a function? e.g. 

```
await safelyCloseDuckDb(quack);
```

```
const safelyCloseDuckDb = async (quack) => {
  await quack.close();
  return new Promise((f) => setTimeout(f, DUCKDB_WRITE_TIMEOUT));
};
```

src/controllers/dataset.ts
j-maynard: Nope... Fixed

src/services/cube-handler.ts
j-maynard: Switched to using a function that combines to the two calls.

wheelsandcogs (approved):

PR #193, opened 2025-04-11
SW-640: Select group that owns the dataset

wheelsandcogs:
Various fixes around user roles and permissions.

 * Create dataset route now requires an id for the group that owns the dataset
 * The JWT now includes the user's groups and roles for the system
 * If a user's perms change, JWT auth will fail and the user will be required to log in again
 * Updated google auth to keep it in sync with EntraID behaviour


j-maynard (approved):
LGTM 🤘🏻

PR #195, opened 2025-04-14
Remove languague checks

j-maynard:


src/utils/lookup-table-utils.ts
```diff
@@ -170,16 +170,18 @@ export const validateLookupTableLanguages = async (
...
-      GROUP BY description, "${joinColumn}" HAVING lang_count > 1
-    `);
-    const duplicateNoteRows = await quack.all(`
-      SELECT "${joinColumn}", notes, COUNT(language) as lang_count
-      FROM (SELECT * FROM "${lookupTableName}" WHERE notes IS NOT NULL)
-      GROUP BY notes, "${joinColumn}" HAVING lang_count > 1
-    `);
+    const duplicateDescriptionRows: TableData[] = [];
+    const duplicateNoteRows: TableData[] = [];
+    // const duplicateDescriptionRows = await quack.all(`
```
j-maynard, line 175: Intensionally commented out while I consider alternative approaches to replace this code.

src/controllers/revision.ts
```diff
@@ -570,13 +575,17 @@ export const downloadRevisionCubeAsJSON = async (req: Request, res: Response, ne
...
-  if (revision.onlineCubeFilename) {
+  if (revision.onlineCubeFilename && !revision.onlineCubeFilename.includes('proto')) {
     const fileBuffer = await req.fileService.loadBuffer(revision.onlineCubeFilename, dataset.id);
     cubeFile = tmp.tmpNameSync({ postfix: '.duckdb' });
     fs.writeFileSync(cubeFile, fileBuffer);
+  } else if (revision.onlineCubeFilename && revision.onlineCubeFilename.includes('proto')) {
+    const fileBuffer = await req.fileService.loadBuffer(revision.onlineCubeFilename, dataset.id);
+    cubeFile = tmp.tmpNameSync({ postfix: '.duckdb' });
+    fs.writeFileSync(cubeFile, fileBuffer);
+    cubeFile = await createBaseCubeFromProtoCube(dataset.id, revision, cubeFile);
```
wheelsandcogs, line 586: There's a lot of duplicated code in these download handlers already that feels ripe for refactoring, but let's try and avoid adding more :D 

```suggestion
  if (revision.onlineCubeFilename) {
    const fileBuffer = await req.fileService.loadBuffer(revision.onlineCubeFilename, dataset.id);
    cubeFile = tmp.tmpNameSync({ postfix: '.duckdb' });
    fs.writeFileSync(cubeFile, fileBuffer);
    if (revision.onlineCubeFilename.includes('proto')) {
      cubeFile = await createBaseCubeFromProtoCube(dataset.id, revision, cubeFile);
    }
```

same pattern should work for the other 4 below I think?

wheelsandcogs (approved):
Would prefer less dupe code, but otherwise ✅

src/controllers/revision.ts
j-maynard: Yeah... I think I need to refactor the code out once I have sometime to do so.

PR #198, opened 2025-04-15
Implement consumer data view

j-maynard:


src/services/consumer-view.ts
```diff
@@ -0,0 +1,95 @@
...
+  datasetId: string,
+  revision: Revision,
+  lang: string,
+  startRow: number,
+  endRow: number,
+  sortBy?: string
+) => {
+  const { parquetMetadata, parquetQuery } = await import('hyparquet');
+  const fileService = getFileService();
+  const stream = await fileService.loadStream(`${revision.id}_${lang}.parquet`, datasetId);
```
wheelsandcogs, line 19: The fileService/datalake has a `loadBuffer()` method - can't we just use that instead of converting the stream?

src/services/consumer-view.ts
```diff
@@ -0,0 +1,95 @@
...
+  const metadata = parquetMetadata(arrayBuffer);
+  const rows = await parquetQuery({ file: asyncBuffer, orderBy: sortBy, rowStart: startRow, rowEnd: endRow });
+
+  return { metadata, rows };
+};
+
+export const getTableMetadata = async (datasetId: string, revision: Revision, lang: string) => {
+  const { parquetMetadata } = await import('hyparquet');
+  const fileService = getFileService();
+  const stream = await fileService.loadStream(`${revision.id}_${lang}.parquet`, datasetId);
```
wheelsandcogs, line 41: Same as above, can we use `loadBuffer()` ?

j-maynard: Yeah... Switching to buffer worked and saved 6 lines of code

src/routes/consumer.ts
```diff
@@ -42,6 +47,8 @@ consumerRouter.get('/list', listPublishedDatasets);
 // GET /published/:datasetId
 // Returns a published dataset as a json object
 consumerRouter.get('/:dataset_id', loadPublishedDataset(), getPublishedDatasetById);
+//13fe4647-c0a2-469b-a159-7ac5bfece4e9
```
wheelsandcogs, line 50: What's this id?

src/services/consumer-view.ts
```diff
@@ -0,0 +1,85 @@
...
+  const tableHeaders = Object.keys(data.rows[0]);
+  const dataArray = data.rows.map((row) => Object.values(row));
+  const headers: CSVHeader[] = [];
+  for (let i = 0; i < tableHeaders.length; i++) {
+    headers.push({
+      index: i - 1,
+      name: tableHeaders[i],
+      source_type: FactTableColumnType.Unknown
+    });
+  }
```
wheelsandcogs, line 71: is `index: i - 1,` deffo right? first loop it's going to be `-1`? 

Optional array.map() 

```suggestion
  const headers: CSVHeader[] = tableHeaders.map((name, idx) => ({
    index: idx - 1,
    name,
    source_type: FactTableColumnType.Unknown
  })); 
```

src/routes/consumer.ts
j-maynard: Rouge dataset ID while I was debugging and working things out.  Removed.

wheelsandcogs (approved):

PR #209, opened 2025-04-22
Fix login issue with match by email

wheelsandcogs:
On the first login where we don't have an Entra user id yet but we do have an email, I'd missed the successful return of the user so that they actually get logged in and instead it was falling through to the "no matching user" error.

The user could re-attempt login and it would be successful the second time around because we were saving the user id on the first attempt.


j-maynard (approved):
LGTM 🤘🏻

PR #215, opened 2025-04-25
on a new revision, the onlineCubeFilename is null, causing this line …

joefitter:
…to throw

### DIFF
--- a/src/controllers/revision.ts
+++ b/src/controllers/revision.ts
@@ -116,7 +116,6 @@ export const getRevisionPreview = async (req: Request, res: Response, next: Next
   const page_size: number = Number.parseInt(req.query.page_size as string, 10) || DEFAULT_PAGE_SIZE;
 
   let cubeFile: string;
-  console.log(revision);
   if (revision.onlineCubeFilename && !revision.onlineCubeFilename.includes('protocube')) {
     logger.debug('Loading cube from file store for preview');
     cubeFile = tmp.tmpNameSync({ postfix: '.duckdb' });


src/controllers/revision.ts
```diff
@@ -126,7 +127,7 @@ export const getRevisionPreview = async (req: Request, res: Response, next: Next
       logger.error('Something went wrong trying to download file from data lake');
       throw err;
     }
-  } else if (revision.onlineCubeFilename.includes('protocube')) {
+  } else if (revision.onlineCubeFilename && revision.onlineCubeFilename.includes('protocube')) {
```
wheelsandcogs, line 130:
```suggestion
  } else if (revision.onlineCubeFilename?.includes('protocube')) {
```

src/controllers/revision.ts
```diff
@@ -126,7 +126,7 @@ export const getRevisionPreview = async (req: Request, res: Response, next: Next
       logger.error('Something went wrong trying to download file from data lake');
       throw err;
     }
-  } else if (revision.onlineCubeFilename.includes('protocube')) {
+  } else if (revision.onlineCubeFilename?.includes('protocube')) {
```
wheelsandcogs, line 129: @j-maynard is onlineCubeFilename only created once the revision is published?

joefitter: this can be seen on dev when looking at https://dev.stats.cymru/en-GB/publish/0c8164e6-dfb4-493c-8fe5-7154206160f5/cube-preview

j-maynard: Nope...  We create a protocube as soon as we've validated the initial data table.   A protocube only has the fact table in... not any of the lookup tables or views.  It saves us having to try to load in all the data every time we want a preview.

j-maynard: So actually now the only time this should be `null` is if we've failed to load the fact table after the preview.

j-maynard (approved):
LGTM 🚀

PR #225, opened 2025-05-06
revert change to /data route, not ready yet

wheelsandcogs:
/data route is still used by dimensions on the frontend, will remove it in the next perf pass, but for now just revert the change to the /data route.


j-maynard (approved):
LGTM 🤘🏻

PR #230, opened 2025-05-07
SW-704: revision routes performance improvements

wheelsandcogs:


src/entities/dataset/dataset.ts
```diff
@@ -63,6 +63,9 @@ export class Dataset extends BaseEntity {
   @JoinColumn({ name: 'end_revision_id', foreignKeyConstraintName: 'FK_dataset_end_revision_id' })
   endRevision: Revision | null;
 
+  @Column({ name: 'draft_revision_id', type: 'uuid', nullable: true })
+  draftRevisionId?: string;
```
wheelsandcogs, line 67: This is a no-op migration wise, just exposes the draft id so we can send it to the frontend and not always have to hydrate the revision

j-maynard (approved):
LGTM 🤘🏻

PR #241, opened 2025-05-12
Wrap cube builder SQL statements in pgformat

j-maynard:
Until now values have not been properly sanitised and SQL statements in the cube builder haven't been properly formatted.  This add sqlstring package and then wraps the function to make it duckdb compatible.  This work should resolve issues around columns and table names not being properly escaped and formatted for SQL.


src/utils/lookup-table-utils.ts
```diff
@@ -54,11 +56,16 @@ export const languageMatcherCaseStatement = (languageColumn: string | undefined)
   SUPPORTED_LOCALES.map((locale) => {
     const lang = locale.split('-')[0].toLowerCase();
     const tLang = lang;
-    languageMatcher.push(`WHEN LOWER("${languageColumn}") LIKE '${lang}%' THEN '${locale.toLowerCase()}'`);
+    languageMatcher.push(pgformat('WHEN LOWER(%I) LIKE %L THEN %L', languageColumn, `%${lang}%`, locale.toLowerCase()));
```
wheelsandcogs, line 59: Are the param placeholders meant to be unique chars? `%L` is repeated here but you have 3 different params.

src/utils/lookup-table-utils.ts
```diff
@@ -54,11 +56,16 @@ export const languageMatcherCaseStatement = (languageColumn: string | undefined)
...
     SUPPORTED_LOCALES.map((locale) => {
       const lang = locale.split('-')[0].toLowerCase();
       languageMatcher.push(
-        `WHEN LOWER("${languageColumn}") LIKE '%${t(`language.${lang}`, { lng: tLang }).toLowerCase()}%' THEN '${locale.toLowerCase()}'`
+        pgformat(
+          'WHEN LOWER(%I) LIKE %L THEN %L',
```
wheelsandcogs, line 64: Same q as above - repeated placeholder.

src/utils/lookup-table-utils.ts
j-maynard: `%L` is a SQL Litteral/Value and `%I` is a SQL Identifier and `%s%` is a raw string.  The library replaces things in the order it's given.  It works in much the same was `printf`

src/utils/lookup-table-utils.ts
j-maynard: `%L` is a SQL Litteral/Value and` %I` is a SQL Identifier and `%s` is a raw string. The library replaces things in the order it's given. It works in much the same was `printf`

src/utils/lookup-table-utils.ts
wheelsandcogs: 👍 should've just read the docs but I was being lazy :D

src/services/cube-handler.ts
```diff
@@ -364,16 +396,41 @@ export const loadCorrectReferenceDataIntoReferenceDataTable = async (quack: Data
...
   }
 };
 
+async function setupReferenceDataDimension(
+  quack: Database,
+  dimension: Dimension,
+  viewSelectStatementsMap: Map<Locale, string[]>,
+  rawSelectStatementsMap: Map<Locale, string[]>,
+  joinStatements: string[]
+  // orderByStatements: string[]
```
wheelsandcogs, line 417: Let's remove this if it's not used.

src/services/cube-handler.ts
```diff
@@ -452,38 +509,66 @@ export async function createDateDimension(quack: Database, extractor: object | n
...
+    rawSelectStatementsMap.get(locale)?.push(`${dimTable}.description as "${columnName}"`);
+  });
+  joinStatements.push(
+    `LEFT JOIN "${dimTable}" on "${dimTable}"."${factTableColumn.columnName}"=${FACT_TABLE_NAME}."${factTableColumn.columnName}" AND "${dimTable}".language=#LANG#`
+  );
+  orderByStatements.push(`"${dimTable}".sort_order`);
+}
+
+// This is a short version of the validate lookup table code found in the dimension process.
+// This concise version doesn't return any information on why the creation failed.  Just that it failed
```
wheelsandcogs, line 556: What's the reason for two versions? Is this faster? If not, what about just using a single version and ignoring/discarding any info you don't need?

src/utils/lookup-table-utils.ts
j-maynard: Old comment... I removed validation a little while ago... Working on the assumption that to get to this stage we've already validated the lookup table.  So I'll remove the comment

src/services/cube-handler.ts
j-maynard: Removed

src/services/cube-handler.ts
j-maynard: Old comment... I removed validation a little while ago... Working on the assumption that to get to this stage we've already validated the lookup table. So I'll remove the comment

wheelsandcogs (approved):

PR #244, opened 2025-05-16
Add support for multiple reference data dimensions

j-maynard:
My original implementation of the reference data queries assumed only one instance of reference data would be used.  This was short sighted as has been proven out by user testing.  This add support for multiple reference data type dimensions.

### DIFF
--- a/src/services/cube-handler.ts
+++ b/src/services/cube-handler.ts
@@ -418,15 +418,58 @@ async function setupReferenceDataDimension(
   await loadCorrectReferenceDataIntoReferenceDataTable(quack, dimension);
   SUPPORTED_LOCALES.map((locale) => {
     const columnName = dimension.metadata.find((info) => info.language === locale)?.name || dimension.factTableColumn;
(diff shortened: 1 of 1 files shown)


src/services/cube-handler.ts
```diff
@@ -418,16 +418,59 @@ async function setupReferenceDataDimension(
...
+      dimension.factTableColumn,
+      `${makeCubeSafeString(dimension.factTableColumn)}_reference_data`
+    )
+  );
+  joinStatements.push(
+    pgformat(
+      `JOIN reference_data_info AS %I ON %I.item_id=%I.item_id`,
+      `${makeCubeSafeString(dimension.factTableColumn)}_reference_data_info`,
+      `${makeCubeSafeString(dimension.factTableColumn)}_reference_data`,
+      `${makeCubeSafeString(dimension.factTableColumn)}_reference_data_info`
```
wheelsandcogs, line 454: consider using a var for `makeCubeSafeString(dimension.factTableColumn)`

wheelsandcogs (approved):

PR #250, opened 2025-05-21
First steps towards postgres cube queries

j-maynard:
This looks like a huge PR... Its not.  Most of the lines added relate to adding CSVs to give a full cube in the Postgres database for the unit tests to run.

### DIFF
--- a/src/services/measure-handler.ts
+++ b/src/services/measure-handler.ts
@@ -586,7 +586,11 @@ async function getMeasurePreviewWithExtractor(dataset: Dataset, measure: Measure
   const quack = await duckdb();
   try {
     await linkToPostgres(quack, revision.id, false);
-    const query = `SELECT * EXCLUDE(language) FROM measure WHERE language = '${lang.toLowerCase()}' ORDER BY sort_order, reference LIMIT ${sampleSize};`;
+    const query = pgformat(
+      `SELECT * EXCLUDE(language) FROM measure WHERE language = %L ORDER BY sort_order, reference LIMIT %Lga ;`,
+      lang.toLowerCase(),
+      sampleSize
+    );
     logger.debug(`Querying the cube to get the preview using query: ${query}`);
     const measureTable = await quack.all(query);
     const tableHeaders = Object.keys(measureTable[0]);
--- a/src/controllers/dimension-controller.ts
+++ b/src/controllers/dimension-controller.ts
@@ -48,7 +48,7 @@ export const resetDimension = async (req: Request, res: Response) => {
       const fileService = getFileService();
       await fileService.delete(lookupTable.filename, dimension.dataset.id);
     } catch (err) {
-      logger.warn(`Something went wrong trying to remove previously uploaded lookup table with error: ${err}`);
+      logger.warn(err, `Something went wrong trying to remove previously uploaded lookup table`);
     }
     await lookupTable.remove();
     dimension.lookuptable = null;


src/enums/cube-type.ts
```diff
@@ -0,0 +1,6 @@
+export enum CubeType {
+  DuckDBProtoCube = 'duckdb_proto_cube',
+  DuckDBCube = 'duckdb_cube',
+  PostgresProtoCube = 'postgres_proto_cube',
+  PostgresCube = 'postgres_cube'
```
wheelsandcogs, line 5: Will the duckdb cubes eventually go away?

src/middleware/services.ts
```diff
@@ -6,7 +6,7 @@ import { getFileService } from '../utils/get-file-service';
 
 // initialise any request-scoped services required by the app and store them on the request object for later use
 // see @types/express/index.d.ts for details
-export const initServices = (req: Request, res: Response, next: NextFunction): void => {
+export const initServices = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
```
wheelsandcogs, line 9: This looks like you might've tried adding the pool to the request object already and then reverted maybe?

src/services/csv-processor.ts
```diff
@@ -13,57 +14,61 @@ import { DatasetRepository } from '../repositories/dataset';
...
     case FileType.Json:
     case FileType.GzipJson:
-      createTableQuery = `CREATE TABLE ${tableName} AS SELECT * FROM read_json_auto('${tempFile}');`;
+      createTableQuery = `CREATE TABLE %I AS SELECT * FROM read_json_auto(%L);`;
       break;
     case FileType.Excel:
       await quack.exec('INSTALL spatial;');
       await quack.exec('LOAD spatial;');
-      createTableQuery = `CREATE TABLE ${tableName} AS SELECT * FROM st_read('${tempFile}');`;
+      createTableQuery = `CREATE TABLE %I AS SELECT * FROM st_read(%L);`;
```
wheelsandcogs, line 58: 👍

src/services/csv-processor.ts
```diff
@@ -73,18 +78,43 @@ export async function extractTableInformation(fileBuffer: Buffer, fileType: File
       `Unknown error occurred, please refer to the log for more information`,
       FileValidationErrorType.unknown
     );
+  }
+
+  if (type === 'data_table')
```
wheelsandcogs, line 83: Can we have some braces here please... my brain can deal with no braces if it's a one-liner but not for a 10 line block 😆

src/services/dimension-processor.ts
```diff
@@ -660,33 +668,35 @@ export const createAndValidateDateDimension = async (
...
+    tableName,
+    tableName,
+    tableName,
+    tableName,
+    dataset.draftRevision!.id,
+    tableName,
+    factTableColumn,
+    tableName,
+    factTableColumn,
+    sampleSize
```
wheelsandcogs, line 697: This feels a bit awkward to use and possibly prone to mistakes - we might want to think about creating a helper function that maps a single copy of each param to the necessary string format for `pgformat`. Not a blocker for this PR though.

src/services/dimension-processor.ts
```diff
@@ -787,24 +797,6 @@ async function getLookupPreviewWithExtractor(
...
-    return viewErrorGenerators(500, dataset.id, 'patch', 'errors.dimension_validation.fact_table_column_not_found', {
-      mismatch: false
-    });
-  }
-
-  logger.debug(`Generating lookup table preview for dimension ${dimension.id}`);
-  const lookupTmpFile = await getFileImportAndSaveToDisk(dataset, dimension.lookupTable);
-  const lookupTableName = `lookup_table`;
-  await loadFileIntoDatabase(quack, dimension.lookupTable, lookupTmpFile, lookupTableName);
-  await createLookupTableInCube(quack, factTableColumn, dimension, lookupTableName);
```
wheelsandcogs, line 807: Good to see so much code being removed 👍

src/services/measure-handler.ts
```diff
@@ -488,76 +540,80 @@ export const validateMeasureLookupTable = async (
...
+    const dataArray = preview.map((row) => Object.values(row));
+    const currentDataset = await DatasetRepository.getById(dataset.id);
+    const headers: CSVHeader[] = [];
+    for (let i = 0; i < tableHeaders.length; i++) {
+      headers.push({
+        index: i,
+        name: tableHeaders[i],
+        source_type: FactTableColumnType.Unknown
+      });
+    }
```
wheelsandcogs, line 568: We do this same map in at least a dozen places, it should have a helper function by now.

There's `tableDataToViewTable()` but it looks like it has an additional wrapper object.

src/services/measure-handler.ts
```diff
@@ -488,76 +540,80 @@ export const validateMeasureLookupTable = async (
...
-    logger.error(`Something went wrong trying to find the measure column for dataset ${dataset.id}`);
-    return viewErrorGenerators(500, dataset.id, 'patch', 'errors.dataset.measure_not_found', {});
-  }
-  if (!measure.measureTable || measure.measureTable.length === 0) {
-    logger.error(`Something went wrong trying to find the measure table for dataset ${dataset.id}`);
-    return viewErrorGenerators(500, dataset.id, 'patch', 'errors.dataset.measure_not_found', {});
+  const quack = await duckdb();
+  try {
+    await linkToPostgres(quack, revision.id, false);
+    const query = `SELECT * EXCLUDE(language) FROM measure WHERE language = '${lang.toLowerCase()}' ORDER BY sort_order, reference LIMIT ${sampleSize};`;
```
wheelsandcogs, line 589: Does this need a `pgformat()`?

test/helpers/test-helper.ts
```diff
@@ -153,6 +165,89 @@ const rowRefLookupTable = () => {
...
+  await linkToPostgresDataTables(quack2);
+  const createDataTableSQL = `CREATE TABLE data_tables_db."${dataTableId}"
+                              (
+                                "YearCode"  BIGINT,
+                                "AreaCode"  BIGINT,
+                                "Data"      DOUBLE PRECISION,
+                                "RowRef"    BIGINT,
+                                "Measure"   BIGINT,
+                                "NoteCodes" VARCHAR
+                              );`;
```
wheelsandcogs, line 238: Crazy indentation batman!

src/controllers/cube-controller.ts
```diff
@@ -80,6 +83,60 @@ export const getCubePreview = async (
...
 
+export const getPostgresCubePreview = async (
+  revisionId: string,
+  lang: string,
+  dataset: Dataset,
+  page: number,
+  size: number
+): Promise<ViewDTO | ViewErrDTO> => {
+  try {
+    const totalsQuery = `SELECT count(*) as "totalLines", ceil(count(*)/${size}) as "totalPages" from "${revisionId}".default_view_${lang};`;
```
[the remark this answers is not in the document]
j-maynard: This is ripped out in the follow up PR being replaced with the same method which drives the consumer view.

src/controllers/cube-controller.ts
```diff
@@ -80,6 +83,60 @@ export const getCubePreview = async (
...
+    const totals = await pool.query(totalsQuery);
+    logger.debug(JSON.stringify(totals.rows));
```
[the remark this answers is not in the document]
j-maynard: Same as previous comment... This whole method is gone in the follow up PR.

src/enums/cube-type.ts
j-maynard: They should... Right now they haven't... Also we need to produce the cube file for Download.  Although the code is moving in to place that there should be a lot less DuckDB usage.

src/services/csv-processor.ts
j-maynard: Braces added

src/services/measure-handler.ts
wheelsandcogs: lang comes from the request - that's a potentially unsafe param.

We should probably just run everything through the sanitiser by default, regardless of where the param comes from.

PR #273, opened 2025-06-03
Reduce size of JWT tokens

wheelsandcogs:
We were hitting size limit (4kb) for the JWT cookie when the user is assigned to many (10+) groups.

This strips the token down to it's bear essentials (no group metadata) to keep the size down. Size with 10 groups is now 1kb.

We can potentially improve this further by switching signing algorithm to something more compressed, but that would require updating the way we do the secret.


j-maynard (approved):
LGTM 🤘🏻

PR #275, opened 2025-06-03
Implement filter table and fix issues with measure lookup

j-maynard:
Implements the filter table in the cube which holds a list of all the possible filter options for all the measures and dimensions in the cube along with their hierarchies.  Adds an end point to get the the table as a nested json object covering the hierarchy into a tree.

Extends the cube rebuild end point to also reimport all data tables associated with a revision and populate them in to the data_tables schema within Postgres.

The PR additionally fixes issues with measure validation which came in from the last PR.

### DIFF
--- a/src/controllers/consumer.ts
+++ b/src/controllers/consumer.ts
@@ -69,7 +69,9 @@ export const getPublishedDatasetView = async (req: Request, res: Response) => {
 };
 
 export const getPublishedDatasetFilters = async (req: Request, res: Response) => {
-  const dataset = await PublishedDatasetRepository.getById(res.locals.datasetId, withAll);
+  const dataset = await PublishedDatasetRepository.getById(res.locals.datasetId, {
+    publishedRevision: true
+  });
   const lang = req.language.toLowerCase();
   if (!dataset.publishedRevision) {
     throw new NotFoundException('errors.no_revision');


src/controllers/consumer.ts
```diff
@@ -68,6 +68,17 @@ export const getPublishedDatasetView = async (req: Request, res: Response) => {
   res.json(preview);
 };
 
+export const getPublishedDatasetFilters = async (req: Request, res: Response) => {
+  const dataset = await PublishedDatasetRepository.getById(res.locals.datasetId, withAll);
```
wheelsandcogs, line 72: It looks like

```
const filters = await getFilters(dataset.publishedRevision, lang);
```

only uses the revision id, so the `withAll` is an expensive hydration that we then don't use any of other than the revision id?

```suggestion
  const dataset = await PublishedDatasetRepository.getById(res.locals.datasetId, {
    publishedRevision: true
  });
```

or if you modify `getFilters` to just take an id instead of a revision object, you could just do `getFilters(dataset.publishedRevisionId, lang);` and not fetch any relations at all.

src/controllers/revision.ts
```diff
@@ -558,6 +578,30 @@ export const regenerateRevisionCube = async (req: Request, res: Response, next:
...
+    }
+  };
+  const dataset = await DatasetRepository.getById(datasetId, datasetRelations);
+  const revisionTree = dataset.revisions
+    .map((rev) => {
+      if (rev.id === revision.id) return rev;
+      else if (rev.revisionIndex > -1) return rev;
+      else return undefined;
+    })
+    .filter((rev) => !!rev)
```
wheelsandcogs, line 593: Looks like this filter is redundant if you've got the one below?

wheelsandcogs (approved):
There's 1 fairly easy to fix perf issue (unnecessary withAll hydration) and some cleanup but otherwise looks good.

PR #277, opened 2025-06-04
Make cube pool a singleton and fail early if can't connect

wheelsandcogs:
Moves the cube db pool config to it's own file, and make sure it only creates a single instance.

Tests the cube db connection on startup so the container will fail to start if the cube db can't connect.


j-maynard (approved):
LGTM 🤘🏻

PR #290, opened 2025-06-09
Move to connect and release on postgres

j-maynard:
When using postgres use a single connection for pulling the views and release immediately after use.

### DIFF
--- a/src/services/consumer-view.ts
+++ b/src/services/consumer-view.ts
@@ -85,29 +85,37 @@ export function transformHierarchy(factTableColumn: string, columnName: string, 
 }
 
 export const getFilters = async (revision: Revision, language: string): Promise<FilterTable[]> => {
-  const cubeDB = getCubeDB();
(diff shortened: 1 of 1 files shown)


src/services/consumer-view.ts
```diff
@@ -85,29 +85,37 @@ export function transformHierarchy(factTableColumn: string, columnName: string,
...
+      const data = columnData.get(col);
+      if (!data) {
+        continue;
+      }
+      const hierarchy = transformHierarchy(data[0].fact_table_column, data[0].dimension_name, data);
+      filterData.push(hierarchy);
     }
-    const hierarchy = transformHierarchy(data[0].fact_table_column, data[0].dimension_name, data);
-    filterData.push(hierarchy);
+    connection.release();
```
easternbloc, line 111: why are we releasing twice?

we should only release in the finally block...

easternbloc (commented):
double release!

PR #294, opened 2025-06-09
SW-789: display revision providers on preview page

wheelsandcogs:
When the providers got moved to the revision, they went missing from the preview/ consumer view. This is needed to restore them.


j-maynard (approved):
LGTM 🤘🏻

PR #300, opened 2025-06-10
Couple more fixes for api docs

wheelsandcogs:
Slightly better examples for topics, missed a rename for page/limit on the topic route.


j-maynard (approved):
LGTM 🤘🏻

PR #301, opened 2025-06-11
Remove the "include all" hydration option to prevent massive queries

wheelsandcogs:
The `IncludeAll` relations option was basically a "give me everything including the kitchen sink" for a dataset object graph.

Most of the time we do not need absolutely everything, and on the rare occasions we might, we should fetch it piece by piece instead of relying on TypeORM to generate an efficient query.

This should hopefully reduce pressure on the query builder and might also help with memory.


j-maynard (approved):
LGTM 🤘🏻

PR #303, opened 2025-06-11
Switch multer storage

j-maynard:
The PR updates the way we use Multer to switch to using Disk storage rather than memory storage to reduce sudden peaks in memory usage especially around data tables.

We've also stripped out code relating to character set detection and re-encoding as this required buffers to do the operations.  As an alternative ducked now supports a `latin-1` option which should cover the most common alternative file encodings.  So code was added to to try UTF-8 and then latin-1 and record this against the data table in the database.

A final fix was made to deal with sample sizes when trying to sniff CSV file data types.  The sniff size has been changed to look at the whole file no matter the size.


src/controllers/revision.ts
```diff
@@ -597,9 +616,41 @@ export const regenerateRevisionCube = async (req: Request, res: Response, next:
...
+      mimetype: rev.dataTable!.mimeType,
+      path: tmpFile,
+      fieldname: '',
+      encoding: '',
+      size: 0,
+      stream: new Readable(),
+      destination: '',
+      filename: '',
+      buffer: Buffer.alloc(0)
+    };
```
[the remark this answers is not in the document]
wheelsandcogs: Although having said that, this regen cube process seems to be:

for each revision in a loop...
1. load original datatable filestream from data lake
2. save filestream to tmp file
3. pass tmp file path to `extractTableInformation()`
4. reload the file again from tmp file, but with duckdb and inject into postgres
5. use duckdb to get the column name, index and datatype for all the columns in the file

Couple of questions that may or may not be dumb:
1. This streams change is to avoid having to load the full file into a buffer?
2. Don't we already have this column info saved in the app db?
  a. Do we actually need to do this file dance for previous revisions rather than just fetching the `DataTableDescription` records for the revision?

src/services/lookup-table-handler.ts
```diff
@@ -223,7 +220,7 @@ export const validateLookupTable = async (
   protoLookupTable: DataTable,
   dataset: Dataset,
   dimension: Dimension,
-  buffer: Buffer,
+  file: Express.Multer.File,
```
wheelsandcogs, line 223: If we're only using `file.path` then probably just pass the path string - if we're using multiple props e.g. path/buffer/origfilename etc then pass the file object.

src/services/measure-handler.ts
```diff
@@ -354,7 +351,7 @@ async function createMeasureTable(
 export const validateMeasureLookupTable = async (
   protoLookupTable: DataTable,
   dataset: Dataset,
-  buffer: Buffer,
+  file: Express.Multer.File,
```
wheelsandcogs, line 354: Same here, just pass the path.

src/controllers/revision.ts
```diff
@@ -597,9 +616,41 @@ export const regenerateRevisionCube = async (req: Request, res: Response, next:
...
+    }
     logger.debug(`Recreating datatable ${rev.dataTable?.id} in postgres data_tables database`);
-    const buffer = await req.fileService.loadBuffer(rev.dataTable!.filename, datasetId);
-    await extractTableInformation(buffer, rev.dataTable!, 'data_table');
+    const tmpFile = await asyncTmpName({ postfix: rev.dataTable!.filename.split('.').reverse()[0] });
+    const downloadStream = await req.fileService.loadStream(rev.dataTable!.filename, dataset.id);
+    const writeStream = fs.createWriteStream(tmpFile);
+    const dataTable = rev.dataTable!;
+    const origEncoding = dataTable.encoding;
+    downloadStream.pipe(writeStream);
```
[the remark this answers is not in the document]
j-maynard: Yeah we need to make sure the stream has come down.. Good catch.

src/services/lookup-table-handler.ts
j-maynard: Changed this to be a string

src/services/measure-handler.ts
j-maynard: Changed this as well.

src/controllers/measure-controller.ts
```diff
@@ -77,6 +80,14 @@
     await createAllCubeFiles(dataset.id, dataset.draftRevision!.id);
     logger.error(err, `An error occurred trying to process and upload the file`);
     next(new UnknownException('errors.upload_error'));
+  } finally {
+    const resolvedPath = path.resolve(multerStorageDir, req.file.path);
+    await stat(resolvedPath)
+      .then(async () => {
+        logger.info(`Deleting temporary file: ${resolvedPath}`);
+        return unlink(resolvedPath);
```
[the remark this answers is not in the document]
j-maynard: This is a variation on the code suggested earlier.

PR #304, opened 2025-06-12
Remove unused dependencies

wheelsandcogs:
Some house keeping -

stripped all dependencies from the package.json that we're not actually using.


j-maynard (approved):
LGTM 🤘🏻

PR #307, opened 2025-06-16
Use the correct server URL for each env in the api docs

wheelsandcogs:
Previously we hardcoded the dev URL, but that doesn't work when viewing the docs on other environments.

Instead, use a placeholder when generating the json spec, and then replace it with the correct backendUrl when serving the docs.


j-maynard (approved):
LGTM 🤘🏻

PR #317, opened 2025-06-19
SW-796: Anti-virus scans of uploaded files

wheelsandcogs:


src/services/virus-scanner.ts
```diff
@@ -0,0 +1,133 @@
...
+        logger.warn(`AV Scan complete. File "${filename}" is infected with: "${viruses}", time: ${time}ms`);
+        reject(new BadRequestException('errors.file_upload.infected'));
+        return;
+      }
+
+      logger.info(`AV Scan complete. File "${filename}" is clean, time: ${time}ms`);
+      resolve(tmpFile);
+    });
+  });
+};
```
wheelsandcogs, line 126: This function returns a TempFile object that provides info (path, etc) about the uploaded file location.

We could optionally create a separate AV function that allows us to pass in stream handlers (e.g. a CSV parser and a Transform that writes to Postgres?) instead of writing to a temporary file, but that's a much bigger chunk of work.

tsconfig.spec.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "isolatedModules": true
+  }
```
wheelsandcogs, line 5: This fixes the ts-jest nagging in tests about isolating modules.

PR #320, opened 2025-06-25
Implement filtered and sorted file downloads

j-maynard:


src/services/consumer-view.ts
```diff
@@ -243,3 +269,166 @@ export const createFrontendView = async (
...
+    let totalRows = 0;
+    let worksheet = workbook.addWorksheet(`Sheet-${sheetCount}`);
+    if (rows.length > 0) {
+      worksheet.addRow(Object.keys(rows[0]));
+      while (rows.length > 0) {
+        for (const row of rows) {
+          if (row === null) break;
+          worksheet.addRow(Object.values(row)).commit();
+        }
+        totalRows += CURSOR_ROW_LIMIT;
```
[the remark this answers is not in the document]
j-maynard: Because we want to check that the next batch of rows doesn't exceed the worksheet limit in Excel we need to check for the maximum of the batch size.

src/controllers/revision.ts
```diff
@@ -767,31 +740,16 @@ export const downloadRevisionCubeAsParquet = async (req: Request, res: Response,
...
-  const datasetId: string = res.locals.datasetId;
-  const revision: Revision = res.locals.revision;
-
-  if (!revision) {
-    next(new UnknownException('errors.no_revision'));
-    return;
+  const revision = res.locals.revision;
+  const lang = req.language.split('-')[0];
+  const sortByQuery = req.query.sort_by ? (JSON.parse(req.query.sort_by as string) as SortByInterface[]) : undefined;
+  const filterQuery = req.query.filter ? (JSON.parse(req.query.filter as string) as FilterInterface[]) : undefined;
```
joefitter, line 746: Same question as above

j-maynard: See above ;-)

joefitter (approved):

PR #327, opened 2025-07-04
Fix download format param in api docs

wheelsandcogs:
The format was showing in both the path and as a query param. This fixes it.

Formats available in the docs should now auto-update as we re-introduce other formats (duckdb etc)


j-maynard (approved):
LGTM 🤘🏻

PR #332, opened 2025-07-04
can't use faker it's a dev dep

wheelsandcogs:
`npm run init:ci` was failing on localstack because the final backend image does not include dev dependencies, of which faker is one.

We no-longer use it anyway so just remove it completely.


wheelsandcogs:
Just gonna merge this.

PR #333, opened 2025-07-07
SW-860: don't break on translation import if measure doesn't exist

wheelsandcogs:
Sometimes datasets might not have a measure column


j-maynard (approved):
LGTM 🤘🏻

PR #335, opened 2025-07-07
Allow large files (500MB) to be streamed to clamAV

wheelsandcogs:
I've picked an arbitrary limit of 500MB... I think ClamAV has a hard limit of 4GB but that's only feasible with a lot more memory dedicated to the container (I think we set 3GB?). 1GB might be doable, but haven't tested it.

I don't think it's completely unreasonable to ask publishers to break up datasets larger than 500MB in to several revisions?

The env vars will need adding to our ClamAV containers @easternbloc.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,3 +36,7 @@ services:
     platform: linux/amd64 # no arm64 image currently available, force amd64
     ports:
       - '3310:3310'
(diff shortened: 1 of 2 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #345, opened 2025-07-14
Ensure unique column names in cube builder and fix sorting

j-maynard:
Pair of things fixed here.  The first is column names need to be unique in views.  While we can't stop a user using the same name for dimensions and measures we can at least postfix them if we find duplicates.  This prevents cube builds failing after poor test data is used in the translation files.

The other thing fixed is natural languague sorting is mostly gone. Sorting is now done against a sort column and then the columns natural sort order.  This should result in more logical sorting of data.

### DIFF
--- a/src/services/cube-handler.ts
+++ b/src/services/cube-handler.ts
@@ -448,15 +448,25 @@ async function setupReferenceDataDimension(
   dimension: Dimension,
   viewSelectStatementsMap: Map<Locale, string[]>,
   rawSelectStatementsMap: Map<Locale, string[]>,
+  defaultSortSelectStatementsMap: Map<Locale, string[]>,
+  rawSortSelectStatementsMap: Map<Locale, string[]>,
+  columnNames: Map<Locale, Set<string>>,
   joinStatements: string[]
 ): Promise<void> {
   await loadCorrectReferenceDataIntoReferenceDataTable(connection, dimension);
   const refDataInfo = `${makeCubeSafeString(dimension.factTableColumn)}_reference_data_info`;
   const refDataTbl = `${makeCubeSafeString(dimension.factTableColumn)}_reference_data`;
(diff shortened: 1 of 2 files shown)


src/services/cube-handler.ts
```diff
@@ -1246,6 +1337,9 @@ async function setupMeasures(
     SUPPORTED_LOCALES.map((locale) => {
       const columnName =
         dataset.measure.metadata.find((info) => info.language === locale)?.name || dataset.measure.factTableColumn;
+      columnNames.get(locale)?.add(t('column_headers.data_values', { lng: locale }));
+      // columnNames.get(locale)?.add(t('column_headers.data_description', { lng: locale }));
```
wheelsandcogs, line 1341: Is this staying?

src/services/cube-handler.ts
```diff
@@ -1354,6 +1521,20 @@ async function dateDimensionProcessor(
...
+    // rawSelectStatementsMap
+    //   .get(locale)
+    //   ?.push(
+    //     pgformat("strftime(%I.start_date, '%d/%m/%Y') AS %I", dimTable, t('column_headers.start_date', { lng: locale }))
+    //   );
+    // rawSelectStatementsMap
+    //   .get(locale)
+    //   ?.push(
+    //     pgformat("strftime(%I.end_date, '%d/%m/%Y') AS %I", dimTable, t('column_headers.end_date', { lng: locale }))
+    //   );
```
wheelsandcogs, line 1537: More commented code, is this staying in?

wheelsandcogs (approved):
Couple of places there's commented out code left in, might want to pull it out.

src/services/cube-handler.ts
j-maynard: See above

PR #348, opened 2025-07-15
SW-739: changes to support the new update frequency UI

wheelsandcogs:
Previously, update frequency was a time interval that we added to the publish date to generate the next expected update date.

This proved unpopular so we have changed the UI to allow the user to enter a specific "next update date".


j-maynard (approved):
LGTM 🤘

PR #352, opened 2025-07-16
SW-904: attempt to fix db connection issues

wheelsandcogs:
Moves the cube db config to using a Typeorm datasource. There are still two separate pools, one for the app db and one for the cube db. Both set to 5 connections by default.

Updated DatabaseManager to manage the cube datasource.

Moved to using Typeorm's QueryRunner in most places for the cube, with the exception of a few places where we use Cursor or pg-copy-stream. We might also be able to ditch pgformat and just use the query runner to do escaping now, but that can be a separate bit of work.

Something to be aware of - 
`QueryRunner` returns the rows directly as an array, there is no object with a `rows` prop.

`PoolClient` does return an object with `rows` by default, so we still need to handle it in those cases.


easternbloc (approved):
LGTM

PR #360, opened 2025-07-23
SW-918: only mark datatable complete if at least one dimension exists 

wheelsandcogs:
If you exited the creation journey before assigning sources, when you go back into the dataset tasklist, the data table section is marked as complete, and it was not obvious you need to go back in and assign sources.

This PR makes it so that the section is marked as incomplete if you have not assigned at least one dimension. Once sources are assigned, it will be marked as complete again, but the dimensions will display as "not started" in the tasklist so it will be obvious they need to be completed.

Screenshot 2025-07-23 at 16 58 54


j-maynard (approved):
LGTM 🤘🏻

PR #362, opened 2025-07-24
SW-928: fix totalPages calculation

wheelsandcogs:
The total_pages query was off by 1, so you could never see the last page of the preview.

We can work it out from the total_lines and page size so do that instead.


j-maynard (approved):
LGTM 🤘🏻

PR #369, opened 2025-07-28
SW-926: CubeDB was missing the search path so validation check failed to find fact_table

wheelsandcogs:
CubeDB query passthrough got missed in a couple of places when trying to fix the run away db connections in https://github.com/Marvell-Consulting/statswales-backend/pull/352

The validation was failing to query the fact table because the search path was not set to the correct cube.


j-maynard (approved):
LGTM 🤘🏻

PR #373, opened 2025-07-29
SW-946: clear any existing PK before attempting to assign one

wheelsandcogs:
When attempting to re-assign sources, the fact table already exists in the cube, and attempting to add a PK will fail with the error:

> multiple primary keys for table \"fact_table\" are not allowed

To prevent this, remove any existing PK first so that we're starting from a blank slate.


j-maynard (approved):
LGTM 🤘🏻

PR #383, opened 2025-08-04
SW-973: allow reverted update drafts to be deleted

wheelsandcogs:
Reverted draft updates were not resetting the revision_index back to zero, so the check to prevent deletion of published revisions `(revision.revisionIndex !== 0)` was triggering.

Fix is to set the rev_index back to 0 when withdrawing a draft update.


easternbloc:
LGTM

PR #390, opened 2025-08-06
SW-990: replace given/family name with single name column

wheelsandcogs:
Previously we supported Google Auth that returns separate given and family names.

EntraID only returns a single name field in the profile, and the assumption that all names would be returned as `{givenName} {familyName}` has proved incorrect. Most names from WG EntraID are returned as `{familyName}, {givenName}` but not always.

Instead, just save a single name field and we'll sort the full name alphabetically instead of attempting to sort on family name.


j-maynard (approved):
LGTM 🤘🏻

PR #404, opened 2025-08-21
Update dependencies

wheelsandcogs:
Bumps our dependencies to latest and remove unused stuff.

The only dep that hasn't been bumped is `openid-client` cause that requires a bit of a rewrite of our auth layer, which is better to do as a standalone PR.


j-maynard (approved):
LGTM 🤘

PR #405, opened 2025-08-21
Reduce log noise

wheelsandcogs:
Reduces the wall of text that the backend debug logs have become.

Query logging is now at `trace` level so set `LOG_LEVEL=trace` to enable.


easternbloc (approved):
LGTM!

PR #406, opened 2025-08-21
Refactor auth to support latest openid-connect

wheelsandcogs:
We want to keep on top of openid-client updates to make sure we're getting the latest security patches.

Version 6.x has been out since October 2024 but we've been putting off the major version bump as it required a refactor of our code.

This is that version bump + refactor.


j-maynard (approved):
LGTM 🤘🏻

PR #412, opened 2025-08-26
SW-774: fix current topic being returned in parent list

wheelsandcogs:
Fix: The get parent topics function was returning the current topic in the list of parents.


j-maynard (approved):
LGTM 🤘

PR #420, opened 2025-09-03
include cubeDB conn check in /healthcheck/live

wheelsandcogs:
`/healthcheck/live` route now checks the cube db connection is still responding, and should error if it's down. This will allow Azure to detect a problem with the cube db connection and restart the container automatically when it occurs.


j-maynard (approved):
LGTM 🤘🏻

PR #421, opened 2025-09-03
Developer: search on dataset id

wheelsandcogs:
Allows the developer list to be filtered by partial (from start) or full dataset id.


j-maynard (approved):
LGTM 🤘

PR #450, opened 2025-10-07
Fix measure handler with new DuckDB code

j-maynard:
A hard coded measure table name was in the createMeasureTable function which would have resulted in issues when using a single central duckdb instance.  This now handles processing the measure table file in duckdb as a set of UUIDs.  Also added was clean up code as the code assumes the duckdb isntance would go away when done and all the data with it.


j-maynard:
No yet ready for review... still some issues around the measure SQL generation in duckdb

easternbloc (changes requested):
Didn't mean to approve this.

PR #464, opened 2025-10-13
Add build log entity

j-maynard:
Adds the build log entity and associated enums and migrations in preparation for adding the new the cube builder to the service.

### DIFF
--- a/src/extractors/date-extractor.ts
+++ b/src/extractors/date-extractor.ts
@@ -9,4 +9,6 @@ export interface DateExtractor {
   dateFormat?: string;
   startDay?: number;
   startMonth?: number;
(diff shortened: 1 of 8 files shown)


wheelsandcogs (commented):
Semantically I'd say this feels like more of a job queue than a log - generally logs are immutable, whereas this will be updated once the job is complete?

Currently this is quite cube-building specific, but we might want to consider just having a generic job queue where cube-building is just one of the types of jobs. I can think of other background tasks (clean up scripts, cache refresh/pre-warm, search index building, etc) that it would be useful to have a queue for.

Something to discuss in Slack.

wheelsandcogs (approved):

PR #483, opened 2025-10-28
Add Pivot table support for DuckDB

j-maynard:


src/controllers/consumer.ts
```diff
@@ -292,6 +293,57 @@ export const getPostgresPivotTable = async (req: Request, res: Response, next: N
   }
 };
 
+export const getDuckDBPivotTable = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+  /*
+  #swagger.ignore = true
+   */
+  const view = req.query.view as string;
+  const dataset = await PublishedDatasetRepository.getById(res.locals.datasetId, withAll);
```
wheelsandcogs: Is `withAll` relations needed here?

```ts
 const withAll: FindOptionsRelations<Dataset> = {
  createdBy: true,
  factTable: true,
  dimensions: { metadata: true, lookupTable: true },
  measure: { measureTable: true, lookupTable: true },
  publishedRevision: {
    metadata: true,
    revisionProviders: { provider: true, providerSource: true },
    revisionTopics: { topic: true }
  },
  revisions: true
};
```

but we only seem to use `dataset.publishedRevision` below? I think `getPostgresPivotTable()` is doing similar. Better to only grab what's actually used by the function.

PR #486, opened 2025-10-30
Linting update

wheelsandcogs:
Adds the import rules that we're already using on the frontend repo.

Enforces braces unless the statement is on a single line.


j-maynard (approved):
I guess... Approve 🤘🏻

PR #508, opened 2025-11-19
SW-1083: fix typo when no topic sort order selected

wheelsandcogs:
The column name for the default applied sort needs to match a key of `sortOpts` on line 199, otherwise we end up with an undefined order by clause in the SQL.


j-maynard (approved):
LGTM 🤘🏻

PR #524, opened 2025-11-26
Add base v2 api files

j-maynard:


scripts/generate-docs.ts
```diff
@@ -11,8 +12,11 @@ import { schema } from '../src/routes/consumer/v1/schema';
  *
  * This should run automatically on build, but can also be run manually with `npm run docs:generate`.
  */
-const consumerEndpoints = ['./src/routes/consumer/v1/api.ts'];
+const consumerEndpoints = ['./src/routes/consumer/v1/api.ts', './src/routes/consumer/v2/api.ts'];
```
wheelsandcogs: This looks like it will include the v2 endpoints in the existing docs?

src/controllers/consumer-v2.ts
```diff
@@ -0,0 +1,368 @@
...
+  const publishedRevision = dataset.publishedRevision;
+  const language = req.language as Locale;
+
+  if (!publishedRevision) {
+    next(new NotFoundException('errors.no_revision'));
+    return {
+      dataset,
+      language,
+      errors: true
+    };
```
wheelsandcogs: This is an unusual pattern for an express app - the `apiSetup` function looks like a request middleware signature `(req, res, next)` but then it doesn't function like one because you're calling it inside an existing request handler like a normal function, and then returning an object as well as calling `next()` with an exception... 

I'm not sure of the actual flow here - what happens to the NotFoundException?

It would be better just to throw the exception and then catch it in the calling function.

PR #530, opened 2025-12-03
Welsh lang support, nicer csv when empty resultset

wheelsandcogs:
Allow the reports to be generated in Welsh.

Provide a header row in the csv when the resultset is empty.


j-maynard (approved):
LGTM 🤘🏻

PR #533, opened 2025-12-05
Remove typeorm-seeding dependency

wheelsandcogs:
We're not really using most of it's features, we can do everything we need without it, so let's just take it out.


j-maynard (approved):
LGTM 🤘🏻

PR #538, opened 2026-01-07
Don't use queryRunner if we don't have to

wheelsandcogs:
We don't need to manage query runner connections manually in most cases, we can call `query()` directly on the datasource and let TypeORM handle the pool.


j-maynard (approved):
LGTM 🤘🏻

PR #547, opened 2026-01-19
Add v2 pivot points

j-maynard:
Adds end points for pivoting existing queries as well as creating new pivot queries with validation to check non x/y variables/dimensions have only one value in them.  Pivots can be rendered in JSON, HTML, CSV, Excel as well as Frontend JSON for frontend rendering.

Pivots are created in DuckDB and streamed out as DuckDB now only streams about the first 2048 rows of any pivot.  Paging and page size are both also supported.

### DIFF
--- a/src/controllers/consumer-v2.ts
+++ b/src/controllers/consumer-v2.ts
@@ -169,6 +169,10 @@ export const getPublishedDatasetData = async (req: Request, res: Response, next:
     const query = await buildDataQuery(queryStore, pageOptions);
     await sendFormattedResponse(query, queryStore, pageOptions, res);
   } catch (err) {
+    if (res.headersSent) {
+      logger.error(err, 'Error detected fetching data after headers already sent');
+      return;
+    }
     if (err instanceof NotFoundException || err instanceof BadRequestException) {
       return next(err);
     }
@@ -316,7 +320,7 @@ export const generatePivotFilterId = async (req: Request, res: Response, next: N
 };
 
 export const generateFilterId = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
-  logger.debug(`Generating filter ID for published dataset ${res.locals.datasetId}...`);
+  logger.info(`Generating filter ID for published dataset ${res.locals.datasetId}...`);
   const dataset = res.locals.dataset as Dataset;
   if (!dataset.publishedRevisionId) return next(new NotFoundException('errors.no_published_revision'));
 
@@ -325,6 +329,9 @@ export const generateFilterId = async (req: Request, res: Response, next: NextFu
     const queryStore = await QueryStoreRepository.getByRequest(dataset.id, dataset.publishedRevisionId, dataOptions);
     res.json({ filterId: queryStore.id });
   } catch (err) {
+    if (err instanceof NotFoundException || err instanceof BadRequestException) {
+      return next(err);
+    }
     logger.error(err, 'Error generating filter ID');
     return next(new UnknownException());
   }
--- a/src/db/entity-subscriber.ts
+++ b/src/db/entity-subscriber.ts
@@ -21,7 +21,7 @@ type AnyEntity = BaseEntity | ObjectLiteral;
 type WriteEvent = InsertEvent<AnyEntity> | UpdateEvent<AnyEntity> | RemoveEvent<AnyEntity>;
 
 // prevent logging of event_log table (infinite loop!) and anything else we want to ignore
-const ignoreTables: string[] = ['event_log', 'build_log'];
+const ignoreTables: string[] = ['event_log', 'build_log', 'query_store'];
 
 // ignore some common props from the logged value that can be easily retrieved elsewhere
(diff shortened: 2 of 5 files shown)


src/controllers/consumer-v2.ts
```diff
@@ -127,6 +207,143 @@ export const getPublishedDatasetData = async (req: Request, res: Response, next:
...
+  const dataOptions = await dtoValidator(DataOptionsDTO, req.body);
+  if (!dataOptions.pivot) {
+    throw new BadRequestException('errors.no_pivot_information');
+  }
+  if (!dataOptions.pivot.x) {
+    throw new BadRequestException('errors.pivot_requires_x');
+  }
+  if (!dataOptions.pivot.y) {
+    throw new BadRequestException('errors.pivot_requires_y');
+  }
```
wheelsandcogs: You could extend `DataOptionsDTO` to make the pivot prop non-optional and let `dtoValidator()` do these checks:

```ts
export class PivotOptionsDTO extends DataOptionsDTO {
  @ValidateNested()
  @Type(() => PivotDTO)
  pivot: PivotDTO;
}
```

src/controllers/consumer-v2.ts
```diff
@@ -127,6 +207,143 @@ export const getPublishedDatasetData = async (req: Request, res: Response, next:
...
+      : await QueryStoreRepository.getByRequest(dataset.id, dataset.publishedRevisionId, dataOptions);
+
+    if (!queryStore.requestObject.pivot) {
+      throw new BadRequestException('errors.not_a_pivot_filter');
+    }
+
+    pageOptions.x = queryStore.requestObject.pivot.x;
+    pageOptions.y = queryStore.requestObject.pivot.y;
+
+    await createPivotFromQuery(res, queryStore, pageOptions);
```
wheelsandcogs: This doesn't just create the query, it also sends the response - this was the main pattern I tried to remove in the refactor.

It helps with readability and debugging if the response is sent from the top level action (or at least in a clearly named "send" helper function) rather than nested at the end of a chain of several nested functions used to fetch data.

This could be as simple as returning the stream object from `createPivotFromQuery()` and renaming `formatChooser()` to `sendPivotResponse()` and calling it directly from here with the stream instead of at the end of `createPivotFromQuery()`.

This is cleaner because then `createPivotFromQuery()` has no need of the `res` object itself, and you can re-use it internally in other code that may not even be related to sending a response to a client.

src/controllers/consumer-v2.ts
```diff
@@ -127,6 +207,143 @@ export const getPublishedDatasetData = async (req: Request, res: Response, next:
...
+  } catch (err) {
+    if (err instanceof NotFoundException || err instanceof BadRequestException) {
+      return next(err);
+    }
+    logger.error(err, 'Error getting published dataset data');
+    next(new UnknownException());
+  }
+};
```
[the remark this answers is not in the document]
j-maynard: So `generatePivotFilterId` validates the Pivot post object and generates the underlying query according the rules we laid out around single values in all variables but X and Y along with a single X and Y value.  This will be the published end point and the one the frontend uses.

`getPublishedDatasetPivot` is able to take any existing query and pivot it with support for multidimensional pivots.  This is experimental and needs work but it was the test end point and we can through further improvement maybe make this public futher down the line.

The underlying code for all these things is quite similar.  Create a query and then pivot on that query.  We don't store the pivot query because there isn't a lot of logic or complexity to it unlike the underlying filter query.

src/services/pivots.ts
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  const dataValuesCol = t('column_headers.data_values', { lng: pageOptions.locale });
+
+  let pagingQuery = '';
+  if (pageOptions.pageSize) {
+    pagingQuery = `LIMIT ${pageOptions.pageSize} OFFSET ${pageOptions.pageSize * (pageOptions.pageNumber - 1)}`;
+  }
+
+  let x = pageOptions.x;
```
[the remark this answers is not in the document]
j-maynard: No... Because you can get to this point from either the get request in which case its a query param or from the post request which wouldbe stored in the query store.  For the get on the `:dataset/pivot/:filter_id` endpoint we just add this to the pageOptions object from the queryStore object.  This keeps the code much simpler as we always look for `x` in the same place.

src/controllers/consumer-v2.ts
j-maynard: So I dislike that change... I've not had an issue debugging this generally and I want the certainty of clearing up the duckdb connection if things go wrong.  So I could refactor it so we just pass the query back and then move basically the response handling back in to the controller which would also have to handle the DuckDB connection and directing which type of output we're sending back.

The query creation is 57 lines of code.  The rest of the code is then create the connection, pass the connection with the type to switch and then choose the appropriate output.  The refactor just splits 57 lines in to just another function.

Unlike with filter query creation which has a lot more that can go wrong we just spacialise in checking the X and Y access are present and correct and then update the underlying query to prefix the from clause table with `cube_db.` so we can use the query in DuckDB.

So I can do a refactor but I do question if its really worth doing.

PR #649, opened 2026-04-22
SW-1235 fix cube preview 404 for datasets with a draft update

wheelsandcogs:
## Summary

Fixes SW-1235 — the cube preview page 404s ("You cannot preview this dataset until you have uploaded a data table") as soon as a user starts an update on a published dataset, even if no data has changed.

- `datasetPreview` (`src/controllers/dataset.ts`) resolved the preview against `dataset.endRevisionId`, which points at the just-created draft update revision. Draft revisions are cloned via `RevisionRepository.deepCloneRevision`, which intentionally does not copy `dataTableId` (sharing would break `replaceDataTable`'s delete-on-replace), so the draft has no data table until the user uploads a new file.
- Commit `a2d99aa6` ("404 the preview when no datatable available") then introduced a hard guard that 404s whenever `endRevision.dataTableId` is missing — correct for a brand-new first revision, but wrong while an update is in progress on a dataset that still has a published revision with data.
- Fix: when the draft has no data table of its own, fall back to `dataset.publishedRevisionId` and preview that. Only 404 when neither side has a data table (preserves the original guard for first-revision, pre-upload datasets).

## Test plan

- [x] `npm run check` (format, build, 1011 unit tests) passes locally.
- [x] New unit test: draft `endRevision` with no `dataTableId` + `publishedRevisionId` set → preview proceeds against the published revision id.
- [x] Existing 404 tests kept (renamed for clarity) — no published revision + no data table still returns 404 `errors.no_data_table`.
- [x] Manual: open `/en-GB/publish/<dataset-with-draft-update>/cube-preview` on the publisher — expect the previously-published preview to render instead of the not-found page.
- [ ] Regression: on a brand-new dataset whose first revision has no uploaded data, the same route should still render the not-found / "upload a data table" state.

### DIFF
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -308,14 +308,19 @@ export const datasetPreview = async (req: Request, res: Response, next: NextFunc
 
   try {
     const endRevision = await RevisionRepository.findOneBy({ id: dataset.endRevisionId });
-    if (!endRevision?.dataTableId) return next(new NotFoundException('errors.no_data_table'));
+
+    // A fresh draft update revision has no dataTable of its own until the user uploads a new file;
+    // preview the currently-published revision in that case so update-in-progress datasets still render.
+    const previewRevisionId = endRevision?.dataTableId ? dataset.endRevisionId : dataset.publishedRevisionId;
+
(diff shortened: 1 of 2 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #650, opened 2026-04-22
SW-1235 fallback filters to published revision during draft updates

wheelsandcogs:
## Summary

Follow-up to `#649` (SW-1235 preview fix). The cube-preview page calls `getRevisionFilters(datasetId, endRevisionId)` in the same `Promise.all` as the preview itself, so the same "draft update revision has no cube yet" condition produced an unhandled Postgres `42P01` → 500 on `/dataset/:id/revision/by-id/:revisionId/preview/filters`. With the preview now falling back to the published revision, the filters needed the same treatment to stay consistent and to stop the Promise.all rejecting.

Changes:

- `src/utils/revision.ts` — new `resolvePreviewRevisionId(revision, dataset)` helper encoding the single policy "use the revision's own id if it has a data table, otherwise fall back to `dataset.publishedRevisionId`, otherwise undefined". Comment lives on the helper, not at each call site.
- `src/controllers/revision.ts:149-165` (`getRevisionPreviewFilters`) — uses the helper; returns `errors.no_data_table` only when neither side has a data table.
- `src/controllers/dataset.ts:303-331` (`datasetPreview`) — collapsed onto the same helper, keeping behaviour unchanged.

One named concept, one place to change if the rule shifts later.

## Test plan

- [x] `npm run check` — 57 suites, 1014 tests, build OK.
- [x] New unit tests in `test/unit/controllers/revision.test.ts` for `getRevisionPreviewFilters`: requested revision has data → queries it; draft has no data + published revision exists → queries the published revision; neither has data → 404.
- [x] Frontend e2e (`update-dataset.spec.ts` + `preview-without-data.spec.ts`) pass end-to-end; the `Preview before any changes are made to the update` test remains the active regression guard.

### DIFF
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -70,6 +70,7 @@ import { DataOptionsDTO, DEFAULT_DATA_OPTIONS, FRONTEND_DATA_OPTIONS } from '../
 import { NotFoundException } from '../exceptions/not-found.exception';
 import { QueryStoreRepository } from '../repositories/query-store';
 import { parsePageOptions } from '../utils/parse-page-options';
+import { resolvePreviewRevisionId } from '../utils/revision';
 import { OutputFormats } from '../enums/output-formats';
 import { buildDataQuery, sendCsv, sendExcel, sendFrontendView, sendJson } from '../services/consumer-view-v2';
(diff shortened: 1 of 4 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #671, opened 2026-05-06
SW-1264 Fix possible race condition around cube builder

j-maynard:
The Cube Builder has always been responsible for creating its own build log entities.  However most end points which use the cube builder need to return the build ID back to the frotnend.  There are two places where the actual build log entity itself is needed immediatly after the cube builder has kicked off its build.  To prevent a race condition and sleep patterns the build log entity can now be created outside of the cube builder and passed in instead of passing in an ID when the entity might not have been created.

As the QueryStore clean up logic has now moved to be inside of the cube builder I've additionally removed the logic around doing this when doing cube rebuilds.

Cleaned up the code to no longer use sleep.  AS the cube builder can run synchronously or asynchronous when rebuilding all revisions we can run this synchronously and remove the need to sleep and poll the results of the build.  Also refactored the associated service functions to live inside of `services/dataset.ts`

### DIFF
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -538,8 +538,6 @@ export const updateSources = async (req: Request, res: Response, next: NextFunct
     }
   }
 
-
-
   const updatedDataset = await DatasetRepository.getById(dataset.id);
   const build = await BuildLog.startBuild(revision, CubeBuildType.FullCube, userId);
 
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -716,8 +716,8 @@ export const rebuildDrafts = async (req: Request, res: Response): Promise<void> 
     return;
   }
   const buildLogEntry = await BuildLog.startBuild(null, CubeBuildType.DraftCubes, user.id);
-  res.status(202).json({ build_id: buildLogEntry.id }).end();
   void rebuildDatasetList(buildLogEntry, await RevisionRepository.getAllDraftRevisionIds(), user);
+  res.status(202).json({ build_id: buildLogEntry.id }).end();
 };
 
 export const rebuildAllFilterTables = async (req: Request, res: Response): Promise<void> => {
@@ -736,6 +736,6 @@ export const rebuildAllFilterTables = async (req: Request, res: Response): Promi
     return;
   }
   const buildLogEntry = await BuildLog.startBuild(null, CubeBuildType.AllFilterTables, user.id);
-  res.status(202).json({ build_id: buildLogEntry.id }).end();
   void rebuildAllFilterTablesForRevisions(buildLogEntry, await RevisionRepository.getAllRevisionIds());
+  res.status(202).json({ build_id: buildLogEntry.id }).end();
 };
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -1,5 +1,3 @@
-import { randomUUID } from 'node:crypto';
-
 import { NextFunction, Request, Response } from 'express';
 import { t } from 'i18next';
 import JSZip from 'jszip';
@@ -749,9 +747,9 @@ async function rebuildDatasetList(buildLogEntry: BuildLog, revisionList: Revisio
   };
 
   for (const rev of revisionList) {
-    const buildId = randomUUID();
-    buildScript.all_builds.push(buildId);
-    buildScript.current_build = buildId;
+    const build = await BuildLog.startBuild(rev, CubeBuildType.FullCube, user.id);
+    buildScript.all_builds.push(build.id);
+    buildScript.current_build = build.id;
     buildLogEntry.buildScript = JSON.stringify(buildScript, null, 2);
     buildLogEntry.status = CubeBuildStatus.Building;
     await buildLogEntry.save();
@@ -760,14 +758,12 @@ async function rebuildDatasetList(buildLogEntry: BuildLog, revisionList: Revisio
     } catch (err) {
       logger.warn(err, `[${buildLogEntry.id}]: Failed to rebuild cube for revision ${rev.id}`);
       failedBuilds.push({
-        buildId: buildId.toString(),
+        buildId: build.id.toString(),
         revisionId: rev.id,
         error: JSON.stringify(err)
       });
       continue;
     }
-
-    const build = await BuildLog.startBuild(rev, CubeBuildType.FullCube, user.id);
 
     void createAllCubeFiles(rev.dataset_id, rev.id, user.id, CubeBuildType.FullCube, build).catch((err: Error) => {
       logger.warn(err, 'Cube builder threw an error while trying to rebuild the cube');
@@ -781,15 +777,15 @@ async function rebuildDatasetList(buildLogEntry: BuildLog, revisionList: Revisio
     }
     if (build.status === CubeBuildStatus.Failed) {
       logger.warn(`[${buildLogEntry}]: Cube for revision ${rev.id} has been failed to rebuild.`);
-      buildScript.failed_to_build.push(buildId);
+      buildScript.failed_to_build.push(build.id);
       buildScript.failed_builds++;
       failedBuilds.push({
-        buildId,
+        buildId: build.id,
         revisionId: rev.id,
         error: JSON.stringify(build.errors)
       });
     } else {
-      buildScript.successfully_built.push(buildId);
+      buildScript.successfully_built.push(build.id);
       buildScript.successful_builds++;
       logger.info(`[${buildLogEntry.id}]: Cube for revision ${rev.id} has been rebuilt successfully.`);
     }
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -540,17 +540,11 @@ export const updateSources = async (req: Request, res: Response, next: NextFunct
 
   const updatedDataset = await DatasetRepository.getById(dataset.id);
   const build = await BuildLog.startBuild(revision, CubeBuildType.FullCube, userId);
-
-  try {
-    res.status(202);
-    res.json({
-      dataset: DatasetDTO.fromDataset(updatedDataset),
-      build_id: build.id
-    });
-  } catch (_) {
-    await build.remove();
-  }
-
+  res.status(202);
+  res.json({
+    dataset: DatasetDTO.fromDataset(updatedDataset),
+    build_id: build.id
+  });
   void createAllCubeFiles(updatedDataset.id, revision.id, userId, CubeBuildType.FullCube, build).catch((err) => {
     logger.error(err, `Failed to create cube files for build ${build.id}`);
   });
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -542,15 +542,16 @@ export const updateSources = async (req: Request, res: Response, next: NextFunct
     }
   }
 
-  const buildId = randomUUID();
+  const build = await BuildLog.startBuild(revision, CubeBuildType.FullCube, userId);
+
   const updatedDataset = await DatasetRepository.getById(dataset.id);
   res.json({
     dataset: DatasetDTO.fromDataset(updatedDataset),
-    build_id: buildId
+    build_id: build.id
   });
 
-  void createAllCubeFiles(updatedDataset.id, revision.id, userId, CubeBuildType.FullCube, buildId).catch((err) => {
-    logger.error(err, `Failed to create cube files for build ${buildId}`);
+  void createAllCubeFiles(updatedDataset.id, revision.id, userId, CubeBuildType.FullCube, build).catch((err) => {
+    logger.error(err, `Failed to create cube files for build ${build.id}`);
   });
 };
 
@@ -766,18 +767,14 @@ async function rebuildDatasetList(buildLogEntry: BuildLog, revisionList: Revisio
       continue;
     }
 
-    void createAllCubeFiles(rev.dataset_id, rev.id, user.id, CubeBuildType.FullCube, buildId).catch((err: Error) => {
+    const build = await BuildLog.startBuild(rev, CubeBuildType.FullCube, user.id);
+
+    void createAllCubeFiles(rev.dataset_id, rev.id, user.id, CubeBuildType.FullCube, build).catch((err: Error) => {
       logger.warn(err, 'Cube builder threw an error while trying to rebuild the cube');
     });
 
     await sleep(5000);
-    let build: BuildLog;
-    try {
-      build = await BuildLog.findOneOrFail({ where: { id: buildId.toString() } });
-    } catch (err) {
-      logger.error(err, 'Failed to find build log entry');
-      continue;
-    }
+
     while (!CompleteStatus.includes(build.status)) {
       await sleep(10000);
       await build.reload();
@@ -795,11 +792,6 @@ async function rebuildDatasetList(buildLogEntry: BuildLog, revisionList: Revisio
       buildScript.successfully_built.push(buildId);
       buildScript.successful_builds++;
       logger.info(`[${buildLogEntry.id}]: Cube for revision ${rev.id} has been rebuilt successfully.`);
-      if (buildLogEntry.type === CubeBuildType.DraftCubes) {
-        await QueryStore.delete({ revisionId: rev.id });
-      } else {
-        await QueryStoreRepository.rebuildQueriesForRevision(rev.id);
-      }
     }
     buildScript.current_build = null;
     buildScript.total_builds++;
--- a/src/controllers/dataset.ts
+++ b/src/controllers/dataset.ts
@@ -57,6 +57,7 @@ import { format as pgformat } from '@scaleleap/pg-format/lib/pg-format';
 import { TaskAction } from '../enums/task-action';
 import { TaskService } from '../services/task';
 import { UserGroup } from '../entities/user/user-group';
+import { CubeBuildStatus } from '../enums/cube-build-status';
 import { CubeBuildType } from '../enums/cube-build-type';
 import { BuildLog } from '../entities/dataset/build-log';
 import { RevisionRepository } from '../repositories/revision';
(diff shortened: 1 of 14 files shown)


src/entities/dataset/build-log.ts
```diff
@@ -2,6 +2,7 @@ import { BaseEntity, Column, CreateDateColumn, Entity, JoinColumn, ManyToOne, Pr
 import { Revision } from './revision';
 import { CubeBuildStatus } from '../../enums/cube-build-status';
 import { CubeBuildType } from '../../enums/cube-build-type';
+import { RevisionList } from '../../repositories/revision';
```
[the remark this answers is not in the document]
j-maynard: Removed the import and switched this to use `revision: { id: string } | Revision | null` as suggested

src/controllers/revision.ts
```diff
@@ -459,12 +460,13 @@ export const regenerateRevisionCube = async (req: Request, res: Response, next:
     next(exception);
     return;
```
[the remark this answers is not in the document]
j-maynard: Added build failure code in the catch block.

src/controllers/measure.ts
```diff
@@ -87,24 +87,24 @@ export const attachLookupTableToMeasure = async (req: Request, res: Response, ne
       res.json(result);
       return;
     }
+    build = await BuildLog.startBuild(dataset.draftRevision, CubeBuildType.FullCube, userId);
     result.extension = {
-      build_id: buildID
+      build_id: build.id
     };
     await updateRevisionTasks(dataset, dataset.measure.id, 'measure');
     res.status((result as ViewErrDTO).status || 200);
```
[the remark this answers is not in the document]
j-maynard: Refactored code as suggested.

src/controllers/dataset.ts
```diff
@@ -542,15 +538,16 @@ export const updateSources = async (req: Request, res: Response, next: NextFunct
...
 
-  const buildId = randomUUID();
+  const build = await BuildLog.startBuild(revision, CubeBuildType.FullCube, userId);
+
   const updatedDataset = await DatasetRepository.getById(dataset.id);
   res.json({
     dataset: DatasetDTO.fromDataset(updatedDataset),
-    build_id: buildId
+    build_id: build.id
   });
```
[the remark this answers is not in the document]
j-maynard: Refactored this.

src/controllers/dimension.ts
```diff
@@ -147,23 +147,23 @@ export const attachLookupTableToDimension = async (req: Request, res: Response,
       res.json(result);
       return;
     }
+    build = await BuildLog.startBuild(draftRevision, CubeBuildType.FullCube, userId);
     result.extension = {
-      build_id: buildId
+      build_id: build.id
     };
     await updateRevisionTasks(dataset, dimension.id, 'dimension');
     res.json(result);
```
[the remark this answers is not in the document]
j-maynard: Refactored this

src/controllers/dataset.ts
```diff
@@ -542,15 +538,23 @@ export const updateSources = async (req: Request, res: Response, next: NextFunct
...
+    res.json({
+      dataset: DatasetDTO.fromDataset(updatedDataset),
+      build_id: build.id
+    });
+  } catch (_) {
+    await build.remove();
+  }
+
+  void createAllCubeFiles(updatedDataset.id, revision.id, userId, CubeBuildType.FullCube, build).catch((err) => {
+    logger.error(err, `Failed to create cube files for build ${build.id}`);
```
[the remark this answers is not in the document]
j-maynard: Reverted this change... The likelyhood of writing the response failing feelings unlikely.

src/services/dataset.ts
```diff
@@ -490,3 +493,143 @@ export class DatasetService {
...
+    buildScript.current_build = build.id;
+    buildLogEntry.buildScript = JSON.stringify(buildScript, null, 2);
+    buildLogEntry.status = CubeBuildStatus.Building;
+    await buildLogEntry.save();
+    try {
+      await bootstrapCubeBuildProcess(rev.dataset_id, rev.id);
+    } catch (err) {
+      logger.warn(err, `[${buildLogEntry.id}]: Failed to rebuild cube for revision ${rev.id}`);
+      build.completeBuild(CubeBuildStatus.Failed, undefined, JSON.stringify(err));
+      await build.save();
```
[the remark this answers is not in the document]
j-maynard: Changed this.

test/integration/routes/regenerate-revision-cube.test.ts
```diff
@@ -101,17 +101,17 @@ describe('POST /dataset/:dataset_id/revision/by-id/:revision_id', () => {
...
   test('creates a BuildLog row in the database for the returned build_id', async () => {
     let createAllCubeFilesPromise: Promise<void> | undefined;
 
-    jest.mocked(createAllCubeFiles).mockImplementation((_datasetId, _revisionId, userId, buildType, buildId) => {
+    jest.mocked(createAllCubeFiles).mockImplementation((_datasetId, _revisionId, userId, buildType, _build) => {
       createAllCubeFilesPromise = (async () => {
         const revision = await Revision.findOneBy({ id: _revisionId });
-        await BuildLog.startBuild(revision, buildType!, userId, buildId);
+        await BuildLog.startBuild(revision, buildType!, userId);
       })();
```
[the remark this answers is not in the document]
j-maynard: Changed this to only create the build entry if its not passed in.

src/services/cube-builder.ts
```diff
@@ -84,7 +84,7 @@ export const createAllCubeFiles = async (
     throw new CubeValidationException('Failed to find buildRevision in dataset.');
   }
 
-  const build = await BuildLog.startBuild(buildRevision, buildType, userId, buildId);
+  if (!build) build = await BuildLog.startBuild(buildRevision, buildType, userId);
```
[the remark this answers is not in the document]
j-maynard: I am incline to leave this... If we can't fetch the dataset and revision means we probably can't update the build log status either.

src/controllers/dataset.ts
```diff
@@ -724,94 +720,6 @@ export const rebuildDrafts = async (req: Request, res: Response): Promise<void>
   void rebuildDatasetList(buildLogEntry, await RevisionRepository.getAllDraftRevisionIds(), user);
 };
```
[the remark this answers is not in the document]
j-maynard: Moved the void call to before the sending the response.

src/controllers/dataset.ts
```diff
@@ -831,57 +739,3 @@ export const rebuildAllFilterTables = async (req: Request, res: Response): Promi
   res.status(202).json({ build_id: buildLogEntry.id }).end();
   void rebuildAllFilterTablesForRevisions(buildLogEntry, await RevisionRepository.getAllRevisionIds());
 };
```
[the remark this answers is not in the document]
j-maynard: Moved the void call to before the sending the response.

src/controllers/dataset.ts
```diff
@@ -828,60 +736,6 @@ export const rebuildAllFilterTables = async (req: Request, res: Response): Promi
     return;
   }
   const buildLogEntry = await BuildLog.startBuild(null, CubeBuildType.AllFilterTables, user.id);
-  res.status(202).json({ build_id: buildLogEntry.id }).end();
   void rebuildAllFilterTablesForRevisions(buildLogEntry, await RevisionRepository.getAllRevisionIds());
+  res.status(202).json({ build_id: buildLogEntry.id }).end();
 };
```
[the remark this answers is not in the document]
j-maynard: this change has already been made

src/controllers/dataset.ts
```diff
@@ -720,98 +716,10 @@ export const rebuildDrafts = async (req: Request, res: Response): Promise<void>
     return;
   }
   const buildLogEntry = await BuildLog.startBuild(null, CubeBuildType.DraftCubes, user.id);
-  res.status(202).json({ build_id: buildLogEntry.id }).end();
   void rebuildDatasetList(buildLogEntry, await RevisionRepository.getAllDraftRevisionIds(), user);
+  res.status(202).json({ build_id: buildLogEntry.id }).end();
 };
```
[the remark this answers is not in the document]
j-maynard: this change has already been made

src/services/cube-builder.ts
```diff
@@ -55,8 +55,8 @@ export const createAllCubeFiles = async (
   datasetId: string,
   buildRevisionId: string,
   userId?: string,
-  buildType = CubeBuildType.FullCube,
-  buildId = crypto.randomUUID(),
+  buildType: CubeBuildType = CubeBuildType.FullCube,
```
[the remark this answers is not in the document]
j-maynard: This doesn't cause a typescript error.

src/controllers/dataset.ts
```diff
@@ -720,98 +716,10 @@ export const rebuildDrafts = async (req: Request, res: Response): Promise<void>
     return;
   }
   const buildLogEntry = await BuildLog.startBuild(null, CubeBuildType.DraftCubes, user.id);
-  res.status(202).json({ build_id: buildLogEntry.id }).end();
   void rebuildDatasetList(buildLogEntry, await RevisionRepository.getAllDraftRevisionIds(), user);
+  res.status(202).json({ build_id: buildLogEntry.id }).end();
```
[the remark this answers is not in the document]
j-maynard: Fixed in next commit

src/controllers/dataset.ts
```diff
@@ -828,60 +736,6 @@ export const rebuildAllFilterTables = async (req: Request, res: Response): Promi
     return;
   }
   const buildLogEntry = await BuildLog.startBuild(null, CubeBuildType.AllFilterTables, user.id);
-  res.status(202).json({ build_id: buildLogEntry.id }).end();
   void rebuildAllFilterTablesForRevisions(buildLogEntry, await RevisionRepository.getAllRevisionIds());
+  res.status(202).json({ build_id: buildLogEntry.id }).end();
```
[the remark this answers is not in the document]
j-maynard: Fixed in next commit

src/services/dataset.ts
```diff
@@ -490,3 +493,166 @@ export class DatasetService {
...
+  };
+
+  for (const rev of revisionList) {
+    const build = await BuildLog.startBuild(rev, CubeBuildType.FullCube, user.id);
+    buildScript.all_builds.push(build.id);
+    buildScript.current_build = build.id;
+    buildLogEntry.buildScript = JSON.stringify(buildScript, null, 2);
+    buildLogEntry.status = CubeBuildStatus.Building;
+    await buildLogEntry.save();
+    try {
```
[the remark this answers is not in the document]
j-maynard: Fixed in next commit

src/controllers/dimension.ts
```diff
@@ -147,23 +147,23 @@ export const attachLookupTableToDimension = async (req: Request, res: Response,
...
   } finally {
     void cleanupTmpFile(tmpFile);
   }
 
-  void createAllCubeFiles(dataset.id, dataset.draftRevision!.id, userId, CubeBuildType.FullCube, buildId).catch(
-    (err) => {
-      logger.error(err, 'Something went wrong trying to build the cube after attaching a lookup table');
-    }
-  );
+  void createAllCubeFiles(dataset.id, dataset.draftRevision!.id, userId, CubeBuildType.FullCube, build).catch((err) => {
```
wheelsandcogs: You've already got a `draftRevision` var that has a falsy check on L132 which I think is in scope here? If you use it here instead of `dataset.draftRevision!.id` you can probably drop the non-null assertion `!`

src/controllers/dimension.ts
```diff
@@ -234,21 +234,21 @@ export const updateDimension = async (req: Request, res: Response, next: NextFun
...
     next(new UnknownException('errors.dimension_update'));
+    return;
   }
 
-  void createAllCubeFiles(dataset.id, dataset.draftRevision!.id, userId, CubeBuildType.FullCube, buildId).catch(
-    (err) => {
-      logger.error(err, 'Something went wrong trying to build the cube when updating the dimension');
-    }
-  );
+  void createAllCubeFiles(dataset.id, dataset.draftRevision!.id, userId, CubeBuildType.FullCube, build).catch((err) => {
```
wheelsandcogs: Same deal here, `draftRevision` exists, but it's in the try/catch scope - if you move

```ts
const draftRevision = dataset.draftRevision;

    if (!draftRevision) {
      logger.error('No draft revision found on dataset');
      next(new UnknownException('errors.no_revision'));
      return;
    }
```

above the `try` it should be usable here.

src/controllers/dimension.ts
```diff
@@ -274,17 +274,15 @@ export const updateDimensionMetadata = async (req: Request, res: Response): Prom
   await metadata.save();
   const updatedDimension = await Dimension.findOneByOrFail({ id: dimension.id });
   await updateRevisionTasks(dataset, dimension.id, 'dimension');
-  const buildId = randomUUID();
+  const build = await BuildLog.startBuild(dataset.draftRevision!, CubeBuildType.FullCube, userId);
```
wheelsandcogs, line 286: I realise this is just a small change but it feels like we should have a draftRevision exists check before this as a failsafe like we do in the other methods?

src/controllers/measure.ts
```diff
@@ -162,19 +163,19 @@ export const updateMeasureMetadata = async (req: Request, res: Response, next: N
...
-  const buildID = randomUUID();
-  await createAllCubeFiles(dataset.id, dataset.draftRevision!.id, userId, CubeBuildType.FullCube, buildID).catch(
-    (err) => {
-      logger.error(
-        err,
-        `Something went wrong trying to build the cube after buildID=${buildID} for datasetId=${dataset.id} and revisionId=${dataset.draftRevision!.id}`
-      );
-    }
-  );
+  const build = await BuildLog.startBuild(dataset.draftRevision!, CubeBuildType.FullCube, userId);
```
wheelsandcogs: Should we be checking draftRevision exists before here?

src/services/dataset.ts
```diff
@@ -490,3 +493,188 @@ export class DatasetService {
     await createAllCubeFiles(dataset.id, dataset.draftRevision!.id, user.id);
   }
 }
+
+export async function rebuildDatasetList(
```
wheelsandcogs: This is called `rebuildDatasetList` but it takes a list of revisions that may or may not be from the same dataset... 
What about `rebuildCubesForRevisions()` and move it to the revisions service?

src/services/dataset.ts
```diff
@@ -490,3 +493,188 @@ export class DatasetService {
...
+    buildLogEntry.completeBuild(
+      CubeBuildStatus.Failed,
+      undefined,
+      err instanceof Error ? (err.stack ?? err.message) : String(err)
+    );
+    await buildLogEntry.save();
+  }
+}
+
+export async function rebuildAllFilterTablesForRevisions(
```
wheelsandcogs: Again, this is dealing with revisions, probably makes more sense to put it in the revision service?

src/controllers/dataset.ts
```diff
@@ -699,9 +696,18 @@ export const rebuildAll = async (req: Request, res: Response): Promise<void> =>
       .end();
     return;
   }
+  const revisionIds = await RevisionRepository.getAllRevisionIds();
   const buildLogEntry = await BuildLog.startBuild(null, CubeBuildType.AllCubes, user.id);
   res.status(202).json({ build_id: buildLogEntry.id }).end();
-  void rebuildDatasetList(buildLogEntry, await RevisionRepository.getAllRevisionIds(), user);
+  rebuildDatasetList(buildLogEntry, revisionIds, user).catch(async (err) => {
```
wheelsandcogs: This is an async function call that returns a promise but there's no `await` or `void`.

src/controllers/dataset.ts
```diff
@@ -719,98 +725,19 @@ export const rebuildDrafts = async (req: Request, res: Response): Promise<void>
...
-      if (buildLogEntry.type === CubeBuildType.DraftCubes) {
-        await QueryStore.delete({ revisionId: rev.id });
-      } else {
-        await QueryStoreRepository.rebuildQueriesForRevision(rev.id);
-      }
-    }
-    buildScript.current_build = null;
-    buildScript.total_builds++;
-    buildLogEntry.buildScript = JSON.stringify(buildScript, null, 2);
+  rebuildDatasetList(buildLogEntry, revisionIds, user).catch(async (err) => {
```
wheelsandcogs: As above, this is an async function call that returns a promise but there's no `await` or `void` here.

src/services/dataset.ts
```diff
@@ -490,3 +493,188 @@ export class DatasetService {
...
+        );
+        await build.save();
+        buildScript.failed_to_build.push(build.id);
+        buildScript.failed_builds++;
+        buildScript.current_build = null;
+        buildScript.total_builds++;
+        buildLogEntry.buildScript = JSON.stringify(buildScript, null, 2);
+        await buildLogEntry.save();
+        failedBuilds.push({
+          buildId: build.id.toString(),
```
wheelsandcogs: do we need the `toString()` here? Entity types id as a string?

src/services/dataset.ts
```diff
@@ -490,3 +493,188 @@ export class DatasetService {
...
```
wheelsandcogs: as above, do we need the `toString()` here? BuildLog types id as a string?

src/controllers/dimension.ts
j-maynard: Fixed this

src/controllers/dimension.ts
j-maynard: Moved above the try catch block and updated the call

src/controllers/measure.ts
j-maynard: Added the check for draftRevision and changed this.

src/controllers/dimension.ts
j-maynard: Added the check and changed this to use `draftRevision`

src/services/dataset.ts
j-maynard: Renamed this and moved it to `services/revision`

src/services/dataset.ts
j-maynard: Moved to `services/revision`

src/controllers/dataset.ts
j-maynard: Made this a `void`

src/services/dataset.ts
j-maynard: Removed the `.toString()`

src/services/dataset.ts
j-maynard: Removed the `.toString()`

src/controllers/dataset.ts
j-maynard: Made this a `void`

wheelsandcogs (approved):

PR #680, opened 2026-05-19
SW-1276: Hide archived datasets with an active redirect from taxonomy & search

wheelsandcogs:
## Summary

Hides datasets that are **archived and have an auto-redirect to a replacement** from all consumer-facing dataset listings (SW-1276). These datasets cause confusion: a user picks one in search or the taxonomy, gets auto-redirected to the replacement, and is unsure whether they're looking at the dataset they chose. As more teams adopt the archive feature, these duplicate entries would increasingly pollute search results and the taxonomy.

Archived datasets *without* an auto-redirect remain visible — there is no replacement to send the user to.

## Hide condition

A dataset is hidden from consumer listings when **all** of:

- `archived_at IS NOT NULL`
- `replacement_dataset_id IS NOT NULL`
- `replacement_auto_redirect = TRUE`

## Changes

All in `src/repositories/published-dataset.ts` — the consumer-side repository, shared by both v1 and v2 consumer APIs. A single SQL fragment constant (`HIDE_REDIRECTED_ARCHIVED`) is applied to:

- `listPublishedByLanguage` — the general "list all published datasets" endpoint
- `listPublishedByTopic` — datasets on a topic page
- `getBaseSearchQuery` — shared base for `searchBasic`, `searchBasicSplit`, `searchFTS`, `searchFuzzy`
- `listPublishedTopics` — joins the dataset so a topic does not appear solely because of a hidden dataset

## Tests

Added a `SW-1276` block to `test/integration/repositories/published-dataset.test.ts` (7 tests) covering all four queries: hidden datasets are excluded; archived datasets with no replacement, or with a replacement but no auto-redirect, stay visible; and a topic whose only dataset is hidden drops off the taxonomy.

`npm run check` passes — build, format, and all 1315 tests.

### DIFF
--- a/src/repositories/published-dataset.ts
+++ b/src/repositories/published-dataset.ts
@@ -48,6 +48,11 @@ export const withPublishedRevision: FindOptionsRelations<Dataset> = {
   replacementDataset: { publishedRevision: { metadata: true } }
 };
 
+// a dataset archived with an auto-redirect to a replacement is hidden from all
+// consumer listings — see SW-1276
+const HIDE_REDIRECTED_ARCHIVED =
+  'NOT (d.archived_at IS NOT NULL AND d.replacement_dataset_id IS NOT NULL AND d.replacement_auto_redirect = TRUE)';
+
 export const PublishedDatasetRepository = dataSource.getRepository(Dataset).extend({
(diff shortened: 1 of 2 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #688, opened 2026-05-26
SW-1247: log client IP and rate-limit-bypass validity on backend HTTP requests

wheelsandcogs:
## What

Adds two fields to backend HTTP access logs so future incident investigations can be served from app logs alone, without round-tripping to Azure Front Door's access log:

- `ip` — the Express-resolved client IP, derived from `X-Forwarded-For` via the existing `app.set('trust proxy', 1)`.
- `rateLimitBypass` — boolean indicating whether the request presented a **valid** `x-rate-limit-bypass` token. Reflects validity rather than mere presence so an external caller cannot spoof "ours" in the logs by sending an arbitrary value in that header.

The bypass token value itself is never logged.

Closes [SW-1247](https://marvellconsulting.atlassian.net/browse/SW-1247).

## How

- `src/middleware/rate-limiter.ts` — after the existing `timingSafeEqual` check succeeds, set `req.rateLimitBypass = true` before calling `next()`. The non-matching branch is left untouched and the serializer defaults a missing flag to `false`, so the secret comparison stays in one place.
- `src/utils/logger.ts` — extracted the pino-http `req` serializer into an exported `serializeRequest` function (with a `SerializedRequest` interface) and added `ip` and `rateLimitBypass` to its output. Comment block on the function records the trust-proxy assumption (see "Topology assumption" below).
- `src/@types/express/index.d.ts` — extended the existing `express-serve-static-core` `Request` augmentation with the optional `rateLimitBypass?: boolean` field.

## Tests

- `test/unit/middleware/rate-limiter.test.ts` — all four existing cases now also assert `req.rateLimitBypass` is `true` only when a valid token is presented and `false` (defaulted) for absent / wrong / not-configured token paths.
- `test/unit/utils/logger.test.ts` — new file. Covers serializer output shape, IP propagation, default-false for the bypass flag, IP omission when absent, and that header / body content never leaks into the serialised output even if those fields are present on the request object.

`npm run check` passes (1,450 tests, lint, build).

## Topology assumption

The `ip` field's correctness depends on `app.set('trust proxy', 1)` matching the actual hop count in front of Express.

Verified against `statswales-terraform`: the inbound chain is Client → Azure Front Door → Container Apps ingress → Express, and only Front Door appends to `X-Forwarded-For` (the ACA ingress is transparent internal routing and is IP-locked to Front Door egress ranges in `modules/container_apps/main.tf`). `trust proxy: 1` is therefore correct today.

If a future change introduces another L7 layer in front of the container — e.g. the Application Gateway flagged as "Planned" in `SolutionDesignDocument.md` — `trust proxy` must be bumped to the new hop count or the `ip` field will log the intermediate proxy's address rather than the real client's.

## Privacy note

This adds a second copy of client IPs (already present in Front Door access logs) into the Container Apps console log table. No new processing activity, lawful basis, or DPIA scope — same controller, same data, same purpose, just a second storage location in the same Log Analytics workspace. Worth confirming separately that retention on the app-log table is not longer than retention on the Front Door access-log table.

## Out of scope

- `src/middleware/rate-limiter.ts:18` already logs `req.ip` on 429s ("Rate limit exceeded for IP: …"). Pre-existing; left as-is.
- Production verification that the new fields appear in `ContainerAppConsoleLogs_CL` unmasked — needs to be checked post-deploy in the staging environment.

[SW-1247]: https://marvellconsulting.atlassian.net/browse/SW-1247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/@types/express/index.d.ts
+++ b/src/@types/express/index.d.ts
@@ -9,5 +9,6 @@ declare module 'express-serve-static-core' {
     files?: Internal.Files;
     fileService: StorageService;
     datasetService: DatasetService;
+    rateLimitBypass?: boolean;
   }
 }
--- a/src/middleware/rate-limiter.ts
+++ b/src/middleware/rate-limiter.ts
@@ -36,6 +36,7 @@ export const rateLimiter = (req: Request, res: Response, next: NextFunction): vo
     timingSafeEqual(Buffer.from(bypassToken), Buffer.from(headerToken))
   ) {
     logger.debug('Rate limit bypass token matched, skipping rate limiting for this request.');
+    req.rateLimitBypass = true;
     next();
     return;
   }
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -20,6 +20,36 @@ function createLogger(): pino.Logger {
 
 export const logger = createLogger();
 
+// Exported for unit testing. The `req` passed in is the pino-http-wrapped Express
+// request, so all Express extensions (query, params, augmented fields) are present.
+export interface SerializedRequest {
+  method?: unknown;
+  url?: unknown;
+  query?: unknown;
+  params?: unknown;
+  ip?: unknown;
+  rateLimitBypass: boolean;
+}
+
+export function serializeRequest(
+  req: { rateLimitBypass?: boolean; ip?: string } & Record<string, unknown>
+): SerializedRequest {
+  return {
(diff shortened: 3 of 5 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #689, opened 2026-05-28
Fix double next() in taskAuth middleware causing ERR_HTTP_HEADERS_SENT

wheelsandcogs:
## Summary

Fixes a double-`next()` bug in the `taskAuth` route middleware (`src/routes/task.ts`) that surfaced in prod as `ERR_HTTP_HEADERS_SENT` errors at `taskDecision`.

When an exception was thrown inside the middleware, the `catch` called `next(error)` but the unconditional `next()` after the try/catch still ran. The error middleware sent its response, then the second `next()` invoked the downstream route handler (`taskDecision` / `getTask`), which tried to `res.json(...)` on an already-sent response.

Moved the success `next()` inside the try block so the happy path calls `next()` exactly once and the error path calls `next(error)` exactly once.

## Context

Spotted in prod logs on 2026-05-28 - 5 occurrences of "Cannot set headers after they are sent to the client" with stack trace at `taskDecision` -> `res.json(...)`.

## Test plan

- [ ] `npm run check` passes (lint + build + tests, 1462 tests)
- [ ] Manually exercise `PATCH /task/:task_id` with a task ID that triggers the auth-failure branch (e.g. a user without group access) and confirm the response is a single 403, no `ERR_HTTP_HEADERS_SENT` in logs

### DIFF
--- a/src/routes/task.ts
+++ b/src/routes/task.ts
@@ -38,10 +38,10 @@ const taskAuth = async (req: Request, res: Response, next: NextFunction): Promis
     logger.debug(`User has access to task ${task.id} via dataset ${dataset.id} and group ${dataset.userGroupId}`);
 
     res.locals.task = task;
+    next();
   } catch (error) {
     next(error);
   }
-  next();
 };
 
 // ****** TASK AUTHORISATION MIDDLEWARE ****** //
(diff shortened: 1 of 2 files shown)


j-maynard (approved):
LGTM 🤘🏻

wheelsandcogs:
Added an extra commit to log any unhandled errors so we can work out what's causing them.

Issue #686, opened 2026-05-26
Improve test coverage for under-covered controllers and services


wheelsandcogs:
## Overview
Coverage audit during the closure of #564, #566, #567 and #568 surfaced several files well below the project's 61% statement-coverage threshold that are not named in the existing #563–#568 testing issues. This issue tracks bringing them up to the same 70% statement-coverage target used by the rest of the testing initiative.

## Measured coverage (full `jest --coverage`, 1363 tests passing)

| File | Statements | Branches | Functions | Lines |
|---|---:|---:|---:|---:|
| `src/controllers/admin.ts` | 23.14% | 0% | 0% | 17.97% |
| `src/controllers/build-log.ts` | 34.21% | 0% | 0% | 31.42% |
| `src/controllers/measure.ts` | 19.87% | 0% | 0% | 16.21% |
| `src/controllers/translation.ts` | 21.78% | 0% | 0% | 18.55% |
| `src/services/dataset.ts` | 50.21% | 27.77% | 39.58% | 49.77% |
| `src/services/revision.ts` | 30.85% | 10.12% | 41.66% | 31.29% |
| `src/services/task.ts` | 13.97% | 0% | 5.55% | 14.60% |

Overall project coverage: **75.2% statements / 62.7% branches / 71.3% functions / 74.8% lines.**

## Target Files
- `src/controllers/admin.ts`
- `src/controllers/build-log.ts`
- `src/controllers/measure.ts`
- `src/controllers/translation.ts`
- `src/services/dataset.ts`
- `src/services/revision.ts`
- `src/services/task.ts`

## Tasks
- [x] Add unit tests for `controllers/admin.ts` — admin-only routes (user / group / org management)
- [x] Add unit tests for `controllers/build-log.ts` — build log retrieval and filtering
- [x] Add unit tests for `controllers/measure.ts` — measure CRUD and metadata
- [x] Add unit tests for `controllers/translation.ts` — translation import/export endpoints
- [x] Add unit tests for `services/dataset.ts` — focus on the untested paths (lines 134–199, 259–322, 348–379, 404–411, 425–490)
- [x] Add unit tests for `services/revision.ts` — focus on untested paths (lines 55–296, 307–361, 394–469, 476–546)
- [x] Add unit tests for `services/task.ts` — task lifecycle, status transitions, validation
- [x] Verify each file reaches 70%+ statement coverage

## Acceptance Criteria
[...]

PR #692, opened 2026-06-02
Add unit tests for dataset, revision and task services (#686)

wheelsandcogs:
## Summary

Adds unit tests for three under-covered services flagged in #686, bringing each above the 70% statement-coverage target used across the testing initiative.

| File | Before | After |
|---|---:|---:|
| `src/services/dataset.ts` | 50.2% | **99.6%** |
| `src/services/revision.ts` | 30.9% | **93.5%** |
| `src/services/task.ts` | 14.0% | **100%** |

84 new tests covering both happy and unhappy paths across each service's public methods — task lifecycle and status transitions, revision draft/approval flows, and dataset fact-table replacement.

## Source changes

`src/services/dataset.ts`:
- Switched the draft-revision delete path from `getFileService()` to the injected `this.fileService`, so the path can be exercised in a unit test.
- Kept the `revisionIndex === 1` guard around the dimension/measure reset as a deliberate defensive backstop. The outer guard in `updateFactTable` already throws for any revision other than the first, so the inner check is currently redundant — but it stays as cheap protection against that outer guard being relaxed later, so we never reset a non-first revision's dimensions/measure.

`jest.config.ts`:
- Raised the global coverage thresholds to match the new floor: statements 79, branches 67, functions 75, lines 79. Full-suite coverage now sits at 80.97% / 68.2% / 76.7% / 80.8%.

## Scope

This PR covers the three **services** from #686. The four controllers (`admin`, `build-log`, `measure`, `translation`) are handled separately in #693.

## Verification

- `jest --coverage` — 1551 tests pass, global thresholds met.
- `tsc` build and lint clean.

Part of #686.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -25,10 +25,10 @@ const config: Config = {
   coverageReporters: ['cobertura', 'lcov', 'html', 'text'],
   coverageThreshold: {
     global: {
-      statements: 61,
-      branches: 49,
-      functions: 57,
-      lines: 61
+      statements: 79,
+      branches: 67,
+      functions: 75,
+      lines: 79
     }
   },
   projects: [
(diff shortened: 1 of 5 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #693, opened 2026-06-02
Add unit tests for admin, build-log, measure and translation controllers (#686)

wheelsandcogs:
## Summary

Adds unit tests for the four under-covered controllers flagged in #686, bringing each well above the 70% statement-coverage target, and fixes two bugs the tests surfaced.

| Controller | Before | After |
|---|---:|---:|
| `src/controllers/admin.ts` | 23.1% | **97.5%** |
| `src/controllers/build-log.ts` | 34.2% | **100%** |
| `src/controllers/measure.ts` | 19.9% | **92.4%** |
| `src/controllers/translation.ts` | 21.8% | **92.1%** |

Both happy and unhappy paths are exercised, including the authorization / validation guards (uuid validation, status validation, `by` validation, unique-constraint handling, av-scan failures, missing-measure / missing-revision guards).

## Bugs found and fixed

Writing the tests surfaced two real defects, each fixed here with a one-line `return`. The tests that caught them now pass and act as regression guards.

**Bug 1 — `admin.ts` `loadUserGroup` and `loadUser` called `next()` twice on the not-found path**

```ts
} catch (_err) {
  next(new NotFoundException('errors.no_user_group'));
  return; // <-- added
}
next(); // only runs on the success path now
```

Without the `return` the catch fell through to the unconditional `next()` at the end of the function, invoking `next` twice (once with the error, once without) and letting the request continue to the route handler with nothing loaded in `res.locals` — the same "responds twice" class of bug fixed in `95adb9db`. Now matches the canonical `datasetAuth` middleware (`src/middleware/dataset-auth.ts`).

**Bug 2 — `measure.ts` `updateMeasureMetadata` fell through its missing-measure guard**

```ts
const measure = dataset.measure;
if (!measure) {
  next(new NotFoundException('errors.measure_invalid'));
  return; // <-- added
}
... measure.metadata.find(...) // previously a TypeError on the null measure
```

Without the `return`, execution dereferenced the null `measure` and threw a `TypeError` instead of returning a clean 404. Now matches `getPreviewOfMeasure` / `getMeasureInfo`.

Regression tests:
- `admin.loadUserGroup › calls next exactly once with NotFound when the group cannot be loaded`
- `admin.loadUser › calls next exactly once with NotFound when the user cannot be loaded`
- `measure.updateMeasureMetadata › forwards NotFound and does not throw when the dataset has no measure`

## Scope

This PR covers the four **controllers** from #686. The three services (`dataset`, `revision`, `task`) are handled separately in #692. The `jest.config.ts` coverage thresholds are left unchanged here; raising them is best done once both PRs land.

## Verification

- Per-controller coverage measured via `jest --coverage` (figures above).
- Full suite: **1546 passing, 0 failing**.
- `eslint` and `tsc --noEmit` clean.

Part of #686.

### DIFF
--- a/src/controllers/admin.ts
+++ b/src/controllers/admin.ts
@@ -39,6 +39,7 @@ export const loadUserGroup = async (req: Request, res: Response, next: NextFunct
     res.locals.userGroupId = group.id;
   } catch (_err) {
     next(new NotFoundException('errors.no_user_group'));
+    return;
   }
 
   next();
@@ -57,6 +58,7 @@ export const loadUser = async (req: Request, res: Response, next: NextFunction):
     res.locals.userId = user.id;
   } catch (_err) {
     next(new NotFoundException('errors.no_user'));
+    return;
   }
 
   next();
--- a/src/controllers/measure.ts
+++ b/src/controllers/measure.ts
@@ -142,6 +142,7 @@ export const updateMeasureMetadata = async (req: Request, res: Response, next: N
   const measure = dataset.measure;
   if (!measure) {
     next(new NotFoundException('errors.measure_invalid'));
+    return;
   }
 
   const draftRevision = dataset.draftRevision;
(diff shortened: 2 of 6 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #695, opened 2026-06-04
Audit dependency bump.

wheelsandcogs:
Removes `jest-junit` (wasn't actually used) as it has an audit-flagged `uuid` version.

Overrides the same flagged `uuid` version in `exceljs` with a patched version. This is a forced override as `exceljs` is no-longer maintained, but it should be fine as the api for uuid v4 has not changed.


j-maynard (approved):
LGTM 🤘🏻

PR #698, opened 2026-06-23
SW-1290 Fix dimension validation and column type mismatch on data table update

j-maynard:
Skip lookup table validation for dimension types that don't use lookup tables
(Text, Numeric, Symbol, NoteCodes), preventing false validation errors when
updating a dataset that previously used note codes or custom lookups.

Also alter the fact table column type to TEXT when a data table column's
datatype differs from the existing fact table column, allowing type changes
(e.g. removing a typed lookup and replacing with a plain text column) without
a hard failure.

### DIFF
--- a/src/services/cube-builder.ts
+++ b/src/services/cube-builder.ts
@@ -1023,7 +1023,8 @@ export function dataTableActions(
         dataTableCol?.columnName
       )
     );
-    if (dataTableCol?.columnDatatype !== factTableCol.columnDatatype) {
+    const normalizedDataType = dataTableCol ? normalizeSqlDatatype(dataTableCol.columnDatatype) : undefined;
+    if (normalizedDataType && normalizedDataType !== normalizeSqlDatatype(factTableCol.columnDatatype)) {
       buildStatements.push(
         pgformat('ALTER TABLE %I.%I ALTER COLUMN %I TYPE TEXT;', buildId, FACT_TABLE_NAME, factTableCol.columnName)
       );
--- a/src/services/cube-builder.ts
+++ b/src/services/cube-builder.ts
@@ -1023,6 +1023,11 @@ export function dataTableActions(
         dataTableCol?.columnName
       )
     );
+    if (dataTableCol?.columnDatatype !== factTableCol.columnDatatype) {
+      buildStatements.push(
+        pgformat('ALTER TABLE %I.%I ALTER COLUMN %I TYPE TEXT;', buildId, FACT_TABLE_NAME, factTableCol.columnName)
+      );
+    }
   }
   logger.debug(`Performing action ${dataTable.action} on fact table for data table ${dataTable.id}`);
   switch (dataTable.action) {
--- a/src/services/revision.ts
+++ b/src/services/revision.ts
@@ -44,6 +44,13 @@ import { BuildLog } from '../entities/dataset/build-log';
 import { User } from '../entities/user/user';
 import { CubeBuildStatus } from '../enums/cube-build-status';
 
+const dimensionTypesNotToValidate = [
+  DimensionType.Text,
+  DimensionType.Numeric,
+  DimensionType.Symbol,
+  DimensionType.NoteCodes
+];
+
 export async function attachUpdateDataTableToRevision(
   datasetId: string,
   revision: Revision,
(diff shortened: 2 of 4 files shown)


src/services/revision.ts
```diff
@@ -44,6 +44,13 @@ import { BuildLog } from '../entities/dataset/build-log';
 import { User } from '../entities/user/user';
 import { CubeBuildStatus } from '../enums/cube-build-status';
 
+const dimensionTypesNotToValidate = [
+  DimensionType.Text,
+  DimensionType.Numeric,
+  DimensionType.Symbol,
+  DimensionType.NoteCodes
+];
```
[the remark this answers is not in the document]
j-maynard: Raw only exists during the initial creation process and shouldn't be available after a dataset has been published.

test/unit/services/revision.test.ts
```diff
@@ -313,6 +313,40 @@ describe('revision service', () => {
         attachUpdateDataTableToRevision('ds-1', revision as never, dataTable as never, DataTableAction.Add)
       ).rejects.toThrow('errors.data_table_validation_error');
     });
+
+    const SKIP_TYPES = [DimensionType.Text, DimensionType.Numeric, DimensionType.Symbol, DimensionType.NoteCodes];
```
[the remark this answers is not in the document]
j-maynard: Raw only exists during the initial creation process and shouldn't be available after a dataset has been published.

src/services/cube-builder.ts
```diff
@@ -1023,6 +1023,12 @@ export function dataTableActions(
         dataTableCol?.columnName
       )
     );
+    const normalizedDataType = dataTableCol ? normalizeSqlDatatype(dataTableCol.columnDatatype) : undefined;
+    if (normalizedDataType && normalizedDataType !== normalizeSqlDatatype(factTableCol.columnDatatype)) {
+      buildStatements.push(
+        pgformat('ALTER TABLE %I.%I ALTER COLUMN %I TYPE TEXT;', buildId, FACT_TABLE_NAME, factTableCol.columnName)
```
easternbloc, line 1029: is that always TYPE text?

I don't think you get automatic type cast here...

test/unit/services/cube-builder.test.ts
```diff
@@ -770,6 +770,48 @@ describe('dataTableActions', () => {
...
+      expect(stmts.filter((s) => s.includes('ALTER COLUMN'))).toHaveLength(0);
+    });
+
+    it('places ALTER COLUMN before action statements', () => {
+      const intYearCol = makeCol('year', 0, FactTableColumnType.Dimension, 'INTEGER');
+      const dataTable = makeDataTable(DataTableAction.ReplaceAll, matchingDescriptions);
+      const stmts = dataTableActions(buildId, dataTable, factTableDef, notesCol, dataValuesCol, [intYearCol]);
+      const alterIdx = stmts.findIndex((s) => s.includes('ALTER COLUMN'));
+      const deleteIdx = stmts.findIndex((s) => s.includes('DELETE FROM'));
+      expect(alterIdx).toBeLessThan(deleteIdx);
```
easternbloc: that's a little weird if you ever have -1 returned (no match)

nicer to just:

expect(stmts.some(s => s.includes('DELETE FROM'))).toBe(true)

easternbloc (changes requested):

src/services/revision.ts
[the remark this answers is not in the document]
easternbloc: unless we explicitly guard against `Raw` then why not add it?

src/services/cube-builder.ts
j-maynard: So a `TEXT` field can take any data...  So in this case the user had an existing integer column which they needed to put text data into.  So basically you go for the column type which can take any type of data if the column types don't match.

test/unit/services/cube-builder.test.ts
j-maynard: Changed this in commit [9f66c7f](https://github.com/Marvell-Consulting/statswales-backend/pull/698/commits/9f66c7fbb96a3f3f1c74353ea1eabd4d077d71f3)

src/services/cube-builder.ts
easternbloc: Yeah, what I was leaning toward `ALTER COLUMN ... TYPE TEXT` with no USING clause will fail on a non-text source column so for safety it'd be good to harden against that.

easternbloc (changes requested):

easternbloc (approved):

PR #702, opened 2026-07-14
Improve approval cube-build error handling and observability

easternbloc:
## Improve approval cube-build error handling and observability

Follow-up to the SW-1299 investigation (#701). Two small, self-contained improvements to `DatasetService.approvePublication` — **error-handling and observability only; no change to how cubes are built.**

### Background
While diagnosing SW-1299 we found two gaps in the approval path:

1. **Bootstrap failures weren't recorded.** `bootstrapCubeBuildProcess` runs *before* `createAllCubeFiles` creates the `build_log` row, so a failure during bootstrap (e.g. rebuilding a lookup table) left **no `build_log` entry** — the error only appeared in application logs, making it hard to diagnose from the database.
2. **Publishers saw a raw 500.** `approvePublication` had no `try/catch`, so a cube-build error propagated to the Express error handler with no `status`, producing a 500 whose body was the **raw Postgres error string**.

### Changes
- Open the `BuildLog` row at the start of `approvePublication` (before bootstrap) and pass it into `createAllCubeFiles` (which already accepts an optional `build`). Wrap both the bootstrap and build steps in `try/catch`; on failure, if the row isn't already terminal, complete it as `Failed`. Net effect: a bootstrap-stage failure is now queryable in `build_log` exactly like a build-stage failure — one row per approval, no duplication.
- In that `catch`, log the real error and rethrow an `UnknownException('errors.cube_builder.cube_build_failed')`, so the publisher gets a clean **translated 500** instead of raw Postgres text. The publish/repository steps are left outside the `try` so their errors are unaffected.
- Add the `errors.cube_builder.cube_build_failed` key to `en.json` and `cy.json`.

### Decisions
- **Generic message, not classified.** A failed build shows a single "the cube could not be built — please try again / contact support" message; diagnosis happens via logs + `build_log`. (Classifying common data problems into specific 400 messages was considered and deferred.)

### ⚠️ Needs review
- The **Welsh** translation string is a first pass and should be checked by a Welsh speaker before release.

### Testing
- Unit tests on `approvePublication` (in `test/unit/services/dataset.test.ts`): happy path still publishes; a build failure and a bootstrap failure each mark the `build_log` row `Failed` and throw the generic translated 500 without publishing. Written test-first (they fail on the current code).
- `dataset` service suite (35 tests) and `task` controller suite (20 tests) pass. Lint clean; changed files typecheck.

### DIFF
--- a/src/resources/locales/cy.json
+++ b/src/resources/locales/cy.json
@@ -30,7 +30,7 @@
   "errors": {
     "unknown_error": "Mae gwall anhysbys wedi digwydd",
     "cube_builder": {
-      "cube_build_failed": "Ni ellid adeiladu'r ciwb. Rhowch gynnig arall arni, a chysylltwch â'r gwasanaeth cymorth os yw'r broblem yn parhau."
+      "cube_build_failed": "Ni allai'r ciwb gael ei adeiladu. Ceisiwch eto, a chysylltwch â chymorth os yw'r broblem yn parhau."
     },
     "missing": "Enw ffeil ar Goll",
     "problem": "Mae problem wedi codi",
--- a/src/resources/locales/cy.json
+++ b/src/resources/locales/cy.json
@@ -29,6 +29,9 @@
   },
   "errors": {
     "unknown_error": "Mae gwall anhysbys wedi digwydd",
+    "cube_builder": {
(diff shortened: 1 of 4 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #708, opened 2026-07-15
SW-1304 Fix SQL injection in measure lookup table matcher

j-maynard:
createMeasureTable built DuckDB identifiers by interpolating user-supplied column names into raw template literals before passing them through pgformat's unescaped %s, letting an authenticated publisher break out of the quoting and inject arbitrary SQL (CWE-89). All column identifiers now go through pgformat's %I, and any matcher-supplied column name is checked against the uploaded file's actual headers before use. The measure lookup patch body is also now validated with a class-validator DTO via dtoValidator instead of an unchecked cast.

### DIFF
--- a/src/controllers/measure.ts
+++ b/src/controllers/measure.ts
@@ -8,6 +8,8 @@ import { DatasetDTO } from '../dtos/dataset-dto';
 import { MeasureLookupPatchDTO } from '../dtos/measure-lookup-patch-dto';
 import { NotFoundException } from '../exceptions/not-found.exception';
 import { UnknownException } from '../exceptions/unknown.exception';
+import { BadRequestException } from '../exceptions/bad-request.exception';
+import { dtoValidator } from '../validators/dto-validator';
 import { getMeasurePreview, validateMeasureLookupTable } from '../services/measure-handler';
 import { validateAndUpload } from '../services/incoming-file-processor';
 import { DimensionMetadataDTO } from '../dtos/dimension-metadata-dto';
@@ -79,7 +81,7 @@ export const attachLookupTableToMeasure = async (req: Request, res: Response, ne
   try {
     const dataTable = await validateAndUpload(tmpFile, dataset.id, 'lookup_table');
     const lang = req.language.toLowerCase();
-    const tableMatcher = req.body as MeasureLookupPatchDTO;
+    const tableMatcher = await dtoValidator(MeasureLookupPatchDTO, req.body);
     const result = await validateMeasureLookupTable(dataTable, dataset, tmpFile.path, lang, tableMatcher);
     if ((result as ViewErrDTO).status) {
       const error = result as ViewErrDTO;
@@ -100,6 +102,10 @@ export const attachLookupTableToMeasure = async (req: Request, res: Response, ne
     res.status((result as ViewErrDTO).status || 200);
     res.json(result);
   } catch (err) {
+    if (err instanceof BadRequestException) {
+      next(err);
+      return;
+    }
     logger.error(err, `An error occurred trying to process and upload the file`);
     next(new UnknownException('errors.upload_error'));
     return;
--- a/src/controllers/measure.ts
+++ b/src/controllers/measure.ts
@@ -79,9 +79,9 @@ export const attachLookupTableToMeasure = async (req: Request, res: Response, ne
   let build: BuildLog;
 
   try {
+    const tableMatcher = await dtoValidator(MeasureLookupPatchDTO, req.body);
     const dataTable = await validateAndUpload(tmpFile, dataset.id, 'lookup_table');
     const lang = req.language.toLowerCase();
-    const tableMatcher = await dtoValidator(MeasureLookupPatchDTO, req.body);
     const result = await validateMeasureLookupTable(dataTable, dataset, tmpFile.path, lang, tableMatcher);
     if ((result as ViewErrDTO).status) {
       const error = result as ViewErrDTO;
--- a/src/dtos/measure-lookup-patch-dto.ts
+++ b/src/dtos/measure-lookup-patch-dto.ts
@@ -1,10 +1,37 @@
-export interface MeasureLookupPatchDTO {
-  join_column: string;
-  description_columns: string[];
+import { IsArray, IsOptional, IsString } from 'class-validator';
+
+export class MeasureLookupPatchDTO {
+  @IsOptional()
+  @IsString()
+  join_column?: string;
+
+  @IsOptional()
+  @IsArray()
+  @IsString({ each: true })
+  description_columns?: string[];
+
+  @IsOptional()
+  @IsArray()
+  @IsString({ each: true })
   notes_columns?: string[];
+
+  @IsOptional()
+  @IsString()
   sort_column?: string;
-  measure_type_column: string;
-  format_column: string;
-  decimal_column: string;
-  language_column: string;
+
+  @IsOptional()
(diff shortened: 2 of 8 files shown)


src/services/measure-handler.ts
```diff
@@ -101,23 +116,31 @@ function createExtractor(
     logger.debug('Using user supplied table matcher to match columns');
     return {
```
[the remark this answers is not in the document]
j-maynard: Fixed in commit [fcef7e3](https://github.com/Marvell-Consulting/statswales-backend/pull/708/commits/fcef7e321358e72396447b53de9e135e9ce40ba8)

src/services/measure-handler.ts
```diff
@@ -92,32 +92,61 @@ async function cleanUpMeasure(measureId: string): Promise<void> {
   }
 }
 
+// Ensures a user-supplied table matcher column name matches one of the actual uploaded headers before it is
+// allowed anywhere near SQL construction. Without this, an attacker-controlled string with no corresponding
+// header would flow straight into an identifier position (CWE-89, SW-1304).
+function findMatchedColumn(protoLookupTable: DataTable, columnName: string | undefined): string | undefined {
+  if (!columnName) return undefined;
+  const match = protoLookupTable.dataTableDescriptions.find((info) => info.columnName === columnName);
```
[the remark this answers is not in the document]
j-maynard: Fixed in commit [10068e1](https://github.com/Marvell-Consulting/statswales-backend/pull/708/commits/10068e19165eab6a6daa12114e8fd850259e04bb)

src/services/measure-handler.ts
```diff
@@ -92,32 +92,61 @@ async function cleanUpMeasure(measureId: string): Promise<void> {
...
+      formatColumn: findMatchedColumn(protoLookupTable, tableMatcher?.format_column),
+      decimalColumn: findMatchedColumn(protoLookupTable, tableMatcher?.decimal_column),
+      measureTypeColumn: findMatchedColumn(protoLookupTable, tableMatcher?.measure_type_column),
+      descriptionColumns: tableMatcher.description_columns.map((desc) => {
+        const info = protoLookupTable.dataTableDescriptions.find((info) => info.columnName === desc);
+        if (!info) {
+          throw new FileValidationException(
+            'errors.measure_validation.unknown_matcher_column',
+            FileValidationErrorType.InvalidCsv
+          );
```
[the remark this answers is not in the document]
j-maynard: Fixed in commit [10068e1](https://github.com/Marvell-Consulting/statswales-backend/pull/708/commits/10068e19165eab6a6daa12114e8fd850259e04bb)

src/services/measure-handler.ts
```diff
@@ -477,7 +506,9 @@ export const validateMeasureLookupTable = async (
     extractor = createExtractor(protoLookupTable, tableLanguage, tableMatcher);
   } catch (error) {
     logger.error(error, `Something went wrong trying to create the measure lookup table extractor`);
-    return viewErrorGenerators(400, dataset.id, 'csv', 'errors.measure_validation.no_description_columns', {
+    const errorKey =
+      error instanceof FileValidationException ? error.message : 'errors.measure_validation.no_description_columns';
```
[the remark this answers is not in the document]
j-maynard: Fixed in commit: [a6d72f2](https://github.com/Marvell-Consulting/statswales-backend/pull/708/commits/a6d72f2ebafd617991232a6b2b7fd801db61f444)

src/services/measure-handler.ts
```diff
@@ -241,7 +270,7 @@ async function createMeasureTable(
         if (extractor.notesColumns) {
           const notesCol = extractor.notesColumns.find((col) => col.lang === locale.toLowerCase())?.name;
           if (notesCol) {
-            notesColumnDef = `"${notesCol}"`;
+            notesColumnDef = pgformat('%I', notesCol);
```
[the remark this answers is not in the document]
j-maynard: Fixed in commit [5c3aebb](https://github.com/Marvell-Consulting/statswales-backend/pull/708/commits/5c3aebb41a7d7b1f2ecea54188074fdcc8c8d879)

src/controllers/measure.ts
```diff
@@ -79,7 +81,7 @@ export const attachLookupTableToMeasure = async (req: Request, res: Response, ne
   try {
     const dataTable = await validateAndUpload(tmpFile, dataset.id, 'lookup_table');
     const lang = req.language.toLowerCase();
-    const tableMatcher = req.body as MeasureLookupPatchDTO;
+    const tableMatcher = await dtoValidator(MeasureLookupPatchDTO, req.body);
```
wheelsandcogs: Should we do this check before the validate and upload to fail early?

src/dtos/measure-lookup-patch-dto.ts
```diff
@@ -1,10 +1,37 @@
...
+  @IsString()
+  format_column?: string;
+
+  @IsOptional()
+  @IsString()
+  decimal_column?: string;
+
+  @IsOptional()
+  @IsString()
+  language_column?: string;
```
wheelsandcogs, line 36: Every prop is optional, is that correct?

src/services/measure-handler.ts
```diff
@@ -460,8 +490,12 @@ export const validateMeasureLookupTable = async (
   let confirmedJoinColumn: string | undefined;
   try {
     confirmedJoinColumn = lookForJoinColumn(protoLookupTable, measure.factTableColumn, tableLanguage, tableMatcher);
-  } catch (_err) {
-    return viewErrorGenerators(400, dataset.id, 'csv', 'errors.measure_validation.no_join_column', {
+  } catch (err) {
+    const errorKey =
+      err instanceof Error && err.message === 'errors.measure_validation.unknown_matcher_column'
```
wheelsandcogs, line 495: Should `lookForJoinColumn` return a proper typed exception rather than just an Error?

src/exceptions/validation-exception.ts
```diff
@@ -17,12 +17,14 @@ export enum FileValidationErrorType {
...
   errorTag: string;
-  message: string;
   type: FileValidationErrorType;
   extension: never;
 
   constructor(message: string, type: FileValidationErrorType, status = 400) {
     super(message);
+    // Class field declarations (even untyped ones like `message: string`) are defined immediately after
+    // super() returns, which would otherwise clobber the message Error's own constructor just set.
+    this.message = message;
```
wheelsandcogs: Now we've removed message from the type definition above, isn't this redundant? (i.e. it should be set by the parent Error via `super(message)`?)

src/dtos/measure-lookup-patch-dto.ts
j-maynard: Yes... Because the object itself is optional.  So if there's no matcher object you still want to be able to proceed.

src/controllers/measure.ts
j-maynard: I've reordered this so the validation comes first.

src/services/measure-handler.ts
j-maynard: Its always just returned an error which then just returns an error to the frontend.  If finding the join column and description columns fails this just throws an error.  We catch it here and just return the error.

src/exceptions/validation-exception.ts
j-maynard: Cleaned this up.

wheelsandcogs (approved):

